Issue #14, opened 2014-12-26
Call to undefined function uv_loop_new()


douggr:
Hi,

using `new Amp\UvReactor()` directly (instead of `Amp\getReactor`) doesn't
check if `php-uv` is loaded and then it dies with a 'Call to undefined
function' message.

What about throw a `RuntimeException` if `php-uv` is not loaded?

rdlowrey:
Looks good ... feel free to PR these commits at your leisure.

douggr:
#16

PR #16, opened 2015-01-06
Throws a RuntimeException if php-uv or libevent is not loaded.

douggr:
Using UvReactor or LibeventReactor directly (instead of Amp\getReactor)
doesn't check if php-uv/libevent is loaded and dies with a 'Call to undefined
function' message.


douggr:
closes #14 
Sorry about the last PR. My robot was out of his bits.

PR #17, opened 2015-01-21
Fix waiting for immediates resolution

bwoebi:
If there are immediates generated in immediates, they have to wait until next tick - which may only be triggered in far future or never, depending on if there are other read/writeWatchers or timers…

That means we can sometimes get very weird blocking situations...

### DIFF
--- a/lib/LibeventReactor.php
+++ b/lib/LibeventReactor.php
@@ -66,7 +66,7 @@ class LibeventReactor implements SignalReactor {
             if (empty($this->enabledWatcherCount)) {
                 break;
             }
-            event_base_loop($this->base, EVLOOP_ONCE);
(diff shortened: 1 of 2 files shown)


lib/LibeventReactor.php
```diff
@@ -60,7 +60,7 @@ public function run(callable $onStart = null) {
         }
 
         while ($this->isRunning) {
-            if ($this->immediates && !$this->doImmediates()) {
+            if (!$this->doImmediates()) {
```
staabm: Why did you strip the boolean check here? The func doesnt do anything when immediates are empty?

staabm: After re-reading the PR desc, I got it. You expect new immediates get generated while the loop

bwoebi:
Updated ...

rdlowrey:
:+1:

PR #19, opened 2015-02-05
Updated requirements to PHP 5.5

lyrixx:
as stated in README.md


rdlowrey:
Technically you only *need* 5.4 for it to run ... 5.5 is required only if you want to do concurrency the "correct" way using generators instead of with a synchronous `wait()` call. There are currently some people who run this on 5.4 and as much as it pains me I don't want to require 5.5 until php7 is released.

lyrixx:
Thanks for the reply. I forgot to close this PR. closing it now.

PR #21, opened 2015-02-05
Fixex typos

lyrixx:


lib/SignalReactor.php
```diff
@@ -1,15 +1,15 @@
-<?php
-
-namespace Amp;
-
-interface SignalReactor extends Reactor {
```
lyrixx: I fixed CRLF

lyrixx:
It looks like:

* `Unresolved::watcher` seems useless 
* `UvReactor::garbage` seems useless
* `LibeventReactor::garbage` seems useless

I noticed too there is plenty of "useless" variable like in:

```php
        return $this->startTimer($callback, $msDelay, $msInterval = 0, Watcher::TIMER_ONCE);
```

Is it for readability ? If no, do you want I remove them?

lib/Unresolved.php
```diff
@@ -95,16 +94,4 @@ private function resolve(\Exception $error = null, $result = null) {
             $this->whens = $this->watchers = [];
         }
     }
-
-    private function update($progress) {
```
rdlowrey: This function cannot be removed -- it is used by the rebound closures here: https://github.com/amphp/amp/blob/master/lib/PrivateFuture.php#L22 ...

If the test suite still passes in its absence then I need to add more tests ;)

rdlowrey:
First, thanks for you PR :+1: 

**Addressing your questions ...**

> `Unresolved::watcher` seems useless

This one only appears useless because the `Unresolved` class has closures rebound to it inside the `PrivateFuture::__construct()` method so that only the holder of the `PrivateFuture` instance can resolve the associated promise (which is implemented with an `Unresolved` instance).

> `UvReactor::garbage` seems useless

We have to temporarily store cancelled watchers in the garbage array and periodically remove them otherwise PHP will fatal out if you try to cancel a watcher with a closure callback from inside itself (a valid operation). This is an unfortunate failing on the part of the language that we have to work around. So while it seems like the `$garbage` arrays are useless I assure you that they are not :)

> `LibeventReactor::garbage` seems useless

Same as above. We don't have this problem in the `NativeReactor` due to the differing nature of how the callbacks are referenced from the extensions used with the other reactors.

> I noticed too there is plenty of "useless" variable like in ...

These also actually have reasons for existing. Different things have to happen when watchers are enabled/disabled/cancelled based on the type of event they're watching for. These constants allow us to use a simple `switch` statement to jump directly to the specific action that's required in such instances. 

**So in summary ...**

I will merge in everything that can be merged -- I believe the removal of `Unresolved::update()` is the only thing that can't be kept. Thanks again for the PR!

PR #25, opened 2015-04-27
drop $forceNew param in getReactor()

staabm:
At runtime there are no use cases where one would like to re-create the reactor. only multi instance scenario is unit tests, but those don't need to be covered in the public api.


staabm:
lets see if tests pass ;)

staabm:
.. no related coverage ..

PR #28, opened 2015-05-20
Run travis in 1.0.x branch only for php 5.5/5.6

staabm:


kelunik:
There's no reason not to run it on PHP 7.

staabm:
People with php7 are not supposed to run this version of amp but master instead. Composer takes care of that already.

kelunik:
Other people may use the 1.0.x branch, because they need PHP 5.5 support and their code runs on PHP 5.5 and 7.

rdlowrey:
As @kelunik  mentioned ... a library might require 5.5 ... so while the newer version is preferable if running in php 7 this older branch also needs to work in php 7. The reason for maintaining this compatibility is so that composer users can specify `"^1"` as their amp version constraint and it will give them exactly what they need based on any other dependencies that may require 5.5+

PR #32, opened 2015-08-07
You're going to want to test on 7.0

GrahamCampbell:
In the future, you want 7.0, not 7.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: php
 php:
   - 5.5
   - 5.6
-  - 7
+  - 7.0
 
 before_script:
   - phpenv rehash


bwoebi:
Actually, I think we want amp to run on the newest version, always. In the future, when 7.0 will be out, we'll have `7.0` and `7` I think?

GrahamCampbell:
> Actually, I think we want amp to run on the newest version, always. In the future, when 7.0 will be out, we'll have 7.0 and 7 I think?

But that means in futue you're not testing on 7.0, and only on 5.5, 5.6, 7.1, if you see what i mean.

bwoebi:
Yes, I do. But there's actually no possibility to write >= 5.5 and have it tested on all newer versions.
We'll anyway have to update it for each new PHP minor release.

kelunik:
Yeah, I don't see any value in that change now, but we could merge it, I don't really care. Once PHP 7 is out, we'll probably add `nightly`, but for now that `7` is just what should actually be `nightly`.

GrahamCampbell:
I don't see why you'd care about working on nightly tbh?

GrahamCampbell:
Surely you want to know if it works on current releases?

GrahamCampbell:
7.0 is kind of a current release now given that rc1 is coming in under 2 weeks.

rdlowrey:
I see no harm in merging this as 7.0 is currently an alias of "nightly" ... we can easily add an additional run for nightlies once 7 is officially released if we think it's necessary.

PR #35, opened 2015-09-23
Convert RuntimeException to CombinatorException

kelunik:


staabm:
:+1:

rdlowrey:
What do we think about creating a generic `Amp\Exception extends \RuntimeException` class that we can then subclass? That way it's always easy to tell if something amp-specific went wrong as compared to other things?

The `Amp\CombinatorException` would then descend from `Amp\Exception` in that case.

Thoughts?

/cc @bwoebi

staabm:
@rdlowrey this would also imply that all Amp\Exceptions are RuntimeExceptions.. is this really the case?

rdlowrey:
@staabm The existing php exception hierarchy is stupid anyway. Technically every userland exception is a `RuntimeException` so I'm not sure how much difference it really makes.

bwoebi:
@rdlowrey We have also plenty of LogicExceptions flying around (which according to PHP 7 semantics should be Errors)… So, I wouldn't superclass Amp\Exception with RuntimeException...

rdlowrey:
Yeah -- the problem is I don't want lots of specific exceptions running around that we can't use a catch-all "Amp" exception type to catch. That turns into a nightmare quickly.

bwoebi:
Ultimately, what's the point of having a catch-all amp exception? Is there any use case for these?

Why would you catch TimeoutException and CombinatorException in one same catch-block? They're not related at all, apart from being part of the same project Amp.

rdlowrey:
It's a cognitive load issue to have lots of specific exceptions. I've run into this problem before in other libs of my own (and other people's).

bwoebi:
Totally agree. But the point here is that the exceptions are independent. I'm totally e.g. in favor of making a DNS exception extend a ClientException if it appears as part of a Client request (example from Artax). But timeout and combinators have _nothing_ to do with each other. It might confuse more than help.

rdlowrey:
@bwoebi yeah that makes sense.

@kelunik can you please add unit test(s) to verify that the expected exception type is thrown?

Also, should we consider wrapping uncaught user exceptions that occur inside combinators in the combinator exception?

cspray:
FWIW @rdlowrey I'm in favor of a generic `Amp\Exception`. I feel like it makes it easier to build abstractions over libraries or simply catch everything that a specific library might emit. I know that PHP's internal exception hierarchy is not ideal but I don't think it should stop this library from having its own ideal hierarchy.

staabm:
@cspray I guess we all like having a base Amp\Exception but the problem atm IMO is that we need to subclass it from RuntimeException for BC reasons which doesnt look right to me

PR #38, opened 2016-02-24
Avoid invalid escape sequences.

Danack:
They drive my OCD wild.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -775,7 +775,7 @@ function __coroutineAdvance(CoroutineState $cs) {
         $yielded = $cs->generator->current();
         if (!isset($yielded)) {
             if ($cs->generator->valid()) {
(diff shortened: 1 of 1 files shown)


staabm:
OCD?

Danack:
[OCD](https://en.wikipedia.org/wiki/Obsessive%E2%80%93compulsive_disorder) - I know the code already was working, but I see invalid escape sequences in a string and my left eye starts twitching.

bwoebi:
I still don't know why that is an invalid escape sequence…
It just is a backslash. It's only an escape sequence if followed by a suited character... But meh…

kelunik:
@bwoebi No, it's invalid if followed by non-suited character. Otherwise, new characters like `\u` will be a BC break.

PR #39, opened 2016-02-25
Expose combined exceptions in combinator exception

kelunik:
Current implementation would be a BC break, if we see the constructor as public API, but I don't think our internal exceptions should be public API, well, I mean their constructors.


bwoebi:
Merged via 7046ca47e185e8a87d3d5d4586d40f7ec2e61f55

PR #41, opened 2016-03-20
Add reactor bound state

bwoebi:
It is common to have code like:

https://github.com/amphp/aerys/blob/77994f6319b8b6b8289cf9637e7747bca41257f9/lib/WorkerProcess.php#L23-L29

Or in tests:
https://github.com/amphp/aerys/blob/77994f6319b8b6b8289cf9637e7747bca41257f9/test/ServerTest.php#L408-L410

It is easy to miss reseting or backing up some dependency when swapping reactors and leads to hard to debug bugs. This PR tries to fix that.

Now dependents can just use `$state = \Amp\getState("amp.dns");` or `$filesystem = \Amp\getState("amp.file");`.

In general, I think the single responsibility of the Reactor is managing events, not providing storage; for that reason (and because a new method in the interface would be a major BC break and require a v2 … E_DEPENDENCY_HELL) I'm proposing the implementation the way it is, with a dynamic property.

@rdlowrey @kelunik feedback please :-)

### DIFF
--- a/lib/Reactor.php
+++ b/lib/Reactor.php
@@ -158,4 +158,6 @@ interface Reactor {
      * @return array
      */
     public function info();
+    
+    /* MUST NOT IMPLEMENT __set() */
 }
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -37,6 +37,27 @@ function driver() {
 }
 
 /**
+ * Get reactor bound state by identifier
+ *
+ * @param string $key identifier for state
+ */
+function getState($key) {
+    $reactor = reactor();
+    return isset($reactor->__dynamicState[$key]) ? $reactor->__dynamicState[$key] : null;
+}
+
+/**
+ * Set reactor bound state by identifier to a value
+ *
+ * @param string $key identifier for state
+ * @param $value
+ * @return void
+ */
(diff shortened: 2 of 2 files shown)


kelunik:
Most users won't save their state like that, we still have the same situation like before with the bugs. This is only useful for libraries using global state like `amphp/dns` and `amphp/

rdlowrey:
I will have to look at this more tomorrow but my immediate reaction is to not like the global state. In any case, not ignoring. Will look closer in the morning

bwoebi:
@rdlowrey On that note, just adding that is is meant to replace the actually global state of e.g. amphp/file etc. which directly use `static`; here it will be bound to the reactor. (Still indirectly global because reactor is global, but, oh well… ;-))

rdlowrey:
On the surface of this I don't really feel like managing this state is the reactor's job ... Am I missing something? You could always just wrap the `reactor()` function yourself, right?

kelunik:
@rdlowrey You need to reset things like `amphp/dns` and `amphp/file` every time you swap the reactor.

bwoebi:
Closing as this will be covered by v2.

PR #44, opened 2016-05-07
Add support for react promises in combinators and coroutines

kelunik:
Don't know if we ever want to merge it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
         }
     ],
     "require": {
-        "php": ">=5.5"
+        "php": ">=5.5",
+        "react/promise": "^2.4"
     },
     "require-dev": {
         "phpunit/phpunit": "~4.8",
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -2,6 +2,8 @@
 
 namespace Amp;
 
+use React\Promise\PromiseInterface as ReactPromise;
+
 /**
  * Retrieve the application-wide event reactor instance
  *
@@ -258,6 +260,9 @@ function all(array $promises) {
     foreach ($promises as $key => $promise) {
         if ($promise instanceof Promise) {
             $promise->when($onResolve, [$struct, $key]);
+        } else if ($promise instanceof ReactPromise) {
+            $promise = __transformReactPromise($promise);
+            $promise->when($onResolve, [$struct, $key]);
         } else {
             $struct->results[$key] = $promise;
             if (--$struct->remaining === 0) {
@@ -319,6 +324,9 @@ function some(array $promises) {
     foreach ($promises as $key => $promise) {
         if ($promise instanceof Promise) {
             $promise->when($onResolve, [$struct, $key]);
+        } else if ($promise instanceof ReactPromise) {
+            $promise = __transformReactPromise($promise);
+            $promise->when($onResolve, [$struct, $key]);
         } else {
             $struct->results[$key] = $promise;
             if (--$struct->remaining === 0) {
@@ -364,6 +372,9 @@ function any(array $promises) {
 
     foreach ($promises as $key => $promise) {
         if ($promise instanceof Promise) {
+            $promise->when($onResolve, [$struct, $key]);
+        } else if ($promise instanceof ReactPromise) {
+            $promise = __transformReactPromise($promise);
             $promise->when($onResolve, [$struct, $key]);
         } else {
             $struct->results[$key] = $promise;
@@ -416,6 +427,9 @@ function first(array $promises) {
 
     foreach ($promises as $key => $promise) {
         if ($promise instanceof Promise) {
+            $promise->when($onResolve, [$struct, $key]);
+        } else if ($promise instanceof ReactPromise) {
+            $promise = __transformReactPromise($promise);
             $promise->when($onResolve, [$struct, $key]);
         } else {
             $struct->remaining = 0;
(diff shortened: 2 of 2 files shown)


composer.json
```diff
@@ -18,7 +18,8 @@
         }
     ],
     "require": {
-        "php": ">=5.5"
+        "php": ">=5.5",
+        "react/promise": "^2.4"
```
kelunik: I'd prefer `suggest` here, as it's not a requirement, but then we can't have any version constraints.

lib/functions.php
```diff
@@ -785,7 +812,11 @@ function __coroutineAdvance(CoroutineState $cs) {
                 $result = (PHP_MAJOR_VERSION >= 7) ? $cs->generator->getReturn() : null;
                 $cs->promisor->succeed($result);
             }
-        } elseif ($yielded instanceof Promise) {
+        } elseif ($yielded instanceof Promise || $yielded instanceof ReactPromise) {
+            if ($yielded instanceof ReactPromise) {
+                $yielded = __transformReactPromise($yielded);
+            }
+
```
kelunik: For speed, it's probably better to move it further down and just duplicate the `if` block content.

joshdifabio:
I think this might be better as `if (method_exists($yielded, 'then')) {`, which is what react and guzzle do for interoperability with other promise implementations. That way you don't need to worry about specific promise implementations. It's also what the JS libraries tend to do.

joshdifabio:
I agree. It might also be faster not to create an amp promise but to instead simply to use the existing `then` method.

joshdifabio:
Sorry, that last comment was supposed to be in response to your line comment about duplicating the if-statement (I'm replying from my phone).

kelunik:
I don't think we'll merge this PR anyway. If we add support, that will be by providing another package.

> I agree. It might also be faster not to create an amp promise but to instead simply to use the existing then method.

One could do that, but then you also have to duplicate all the `$onResolve` callbacks.

joshdifabio:
I totally understand that you wouldn't want to add a dependency on react/promise, but what about simply duck-typing to support any thenable [as I suggested](https://github.com/amphp/amp/pull/44#issuecomment-217629185)? This is the approach React and Guzzle both take, as well as most JS promise libraries I believe, and would enable interoperability with React, Guzzle and Icicle promises. Guzzle and React's promise libraries have five million downloads each and masses of dependent projects, surely it's worth supporting them?

composer.json
rdlowrey: Agree with the suggest

bwoebi:
That's doable too. But we probably need an event-loop adapter too. (Else you'll have to tick the loops in a busy main loop like each 50 milliseconds, which is not very nice.)

And at that point, the question really is, whether we shouldn't just introduce proper adapters?

joshdifabio:
> That's doable too. But we probably need an event-loop adapter too. (Else you'll have to tick the loops in a busy main loop like each 50 milliseconds, which is not very nice.)

https://github.com/guzzle/promises/pull/27

If that PR is merged then I believe it'll be possible to run the Guzzle task queue (which is used to invoke promise observers) very efficiently, without the use of timers, and it looks like the same will be possible with react/promise v3. I really think promise interoperability is, and always should be, separate to event loop interoperability. Even if all my IO was handled by the Amp reactor, which is where I'd like to get to, I'd still have masses of code consuming and returning thenables.

rdlowrey:
I know I've been absentee for a while but I might be persuaded to incorporate promise compat separate from event loop compat. It's at least worth considering. Even though I personally think the amp solutions are better (that's why we wrote them that way, after all), we're unlikely to gain wide adoption without finding ways to integrate with these other libs.

It's something to consider.

bwoebi:
@joshdifabio I'm not sure what you're talking about here? The main issue is waiting for I/O.

Typically an event reactor is blocking inside one stream_select() call. That stream_select() call has to know about _all_ the streams we're waiting for.
To allow for another reactors events to be dispatched, you have this reactor to run _his_ stream_select() call with his streams.
The only solutions to this are really calling the other reactors tick function or dispatching the event listeners to our reactor via an adapter.

joshdifabio:
> @joshdifabio I'm not sure what you're talking about here? The main issue is waiting for I/O.

I'm talking about working with a single event loop. React promises and Guzzle promises have precisely zero dependencies on any event loop. It's entirely possible to run only an Amp reactor and still have multiple promise implementations in your application. Does that make sense? Promise interoperability is not dependent on event loop interoperability.

kelunik:
> but what about simply duck-typing to support any thenable as I suggested previously?

That might break other objects having a then method not being a thenable. Technically, even adding the `ReactPromise` is a BC break.

> React promises and Guzzle promises have precisely zero dependencies on any event loop.

The promises don't have that dependency, but the libraries usually have.

Guzzle requires being called in every event loop tick, see https://github.com/guzzle/promises/blob/8c54fb01b3b077f07f74f658eac1c558930cfbc3/README.md#event-loop-integration. That's exactly the busy loop we're talking about.

joshdifabio:
> That might break other objects having a then method not being a thenable. Technically, even adding the ReactPromise is a BC break.

Absolutely any change can be a BC break. When you fix a bug you are changing functionality which people might have been relying on. Sometimes a little pragmatism is helpful.

> Guzzle requires being called in every event loop tick, see https://github.com/guzzle/promises/blob/8c54fb01b3b077f07f74f658eac1c558930cfbc3/README.md#event-loop-integration. That's exactly the busy loop we're talking about.

Which is exactly why I submitted the PR I referenced above, which removes this requirement. If nothing has been added to the task queue then there is no need to run it.

I don't think there's really any more I can say on this, it's up to you guys what your wasn't to do. Thanks for hearing me out.

PR #45, opened 2016-05-11
Make resolve() accept also callables

DaveRandom:
This is just syntactic sugar around `resolve()` for use with a generator closure to avoid temporary variables.

Instead of having to write:

```php
$var = function() {
    yield;
};
return resolve($var);
```

you can simply write:

```php
return promise(function() {
    yield;
});
```

I'm not married to the function name if it doesn't suit, but this is an option I have found frustratingly missing.

This PR also adds a missing `@return` tag to the `resolve()` docblock.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
  * Upon resolution the Generator return value is used to succeed the promised result. If an
  * error occurs during coroutine resolution the returned promise fails.
  *
- * @param \Generator $generator The generator to resolve as a coroutine
+ * @param \Generator|callable $generator A generator or callable the returns a generator to resolve as a coroutine
+ * @return \Amp\Promise
  */
-function resolve(\Generator $generator) {
+function resolve($generator) {
+    if (!$generator instanceof \Generator) {
+        if (is_callable($generator)) {
+            $generator = call_user_func($generator);
+        }
+
+        if (!$generator instanceof \Generator) {
+            throw new \LogicException('Callable passed to resolve() did not return an instance of Generator');
+        }
+    }
+
     $cs = new CoroutineState;
     $cs->promisor = new Deferred;
     $cs->generator = $generator;


joshdifabio:
FYI, if you're able to use PHP7 syntax you can write:

```php
return resolve((function() {
    yield;
})());
```

kelunik:
Another possible solution would be to change `resolve` to accept callables and generators.

DaveRandom:
Yeh I did that at one point but it is indeed not pretty. Also the PHP Storm parser currently doesn't like it, but I believe a fix for that is in the pipeline.

DaveRandom:
@kelunik That's actually a better idea, will update PR to do that instead

bwoebi:
Agree, changing resolve is the best way.

DaveRandom:
Updated

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
...
  */
-function resolve(\Generator $generator) {
+function resolve($generator) {
+    if (is_callable($generator)) {
+        $generator = call_user_func($generator);
+
+        if (!$generator instanceof \Generator) {
+            throw new \LogicException('Callable passed to resolve() did not return a Generator');
+        }
+    } else if (!$generator instanceof \Generator) {
```
joshdifabio: Based on the rest of the code base I think this should be formatted as `elseif`.

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
  * Upon resolution the Generator return value is used to succeed the promised result. If an
  * error occurs during coroutine resolution the returned promise fails.
  *
- * @param \Generator $generator The generator to resolve as a coroutine
+ * @param \Generator|callable $generator The generator to resolve as a coroutine
+ * @return \Amp\Promise
  */
-function resolve(\Generator $generator) {
+function resolve($generator) {
```
staabm: maybe we should rename the var then

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
  * Upon resolution the Generator return value is used to succeed the promised result. If an
  * error occurs during coroutine resolution the returned promise fails.
  *
- * @param \Generator $generator The generator to resolve as a coroutine
+ * @param \Generator|callable $generator The generator to resolve as a coroutine
```
kelunik: Description needs an update.

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
...
-function resolve(\Generator $generator) {
+function resolve($generator) {
+    if (is_callable($generator)) {
+        $generator = call_user_func($generator);
+
+        if (!$generator instanceof \Generator) {
+            throw new \LogicException('Callable passed to resolve() did not return a Generator');
+        }
+    } else if (!$generator instanceof \Generator) {
+        throw new \LogicException('Co-routine to resolve must be callable or Generator');
```
joshdifabio: Wouldn't `InvalidArgumentException` be correct here?

kelunik: I prefer `Coroutine`, don't know which one is right. Also maybe `or instance of Generator`.

lib/functions.php
kelunik: Personally, I always use `else if`. I think it's just the `amp` repo that uses `elseif`, but not sure. @bwoebi @rdlowrey thoughts?

staabm: doesnt our style rules already take care of that?
https://github.com/amphp/amp/blob/master/.php_cs

lib/functions.php
kelunik: @joshdifabio Not sure, a callable might also return a `Generator` conditionally, then the argument is fine, just the return logic not.

joshdifabio: I'm referring to line 766, not 763.

kelunik: Ah, you're right. :+1:

DaveRandom: I agree that should be an `InvalidArgumentException` on L766 (as if the spl extension model makes any sense at all :-P).

WRT coroutine vs co-routine I'll go with the project standard, from an English perspective Co-routine is probably more correct but consistency is more important.

lib/functions.php
bwoebi: I always use elseif ... If you ask me: elseif.

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
  * Upon resolution the Generator return value is used to succeed the promised result. If an
  * error occurs during coroutine resolution the returned promise fails.
  *
- * @param \Generator $generator The generator to resolve as a coroutine
+ * @param \Generator|callable $generator The generator to resolve as a coroutine
+ * @return \Amp\Promise
  */
-function resolve(\Generator $generator) {
+function resolve($generator) {
+    if (is_callable($generator)) {
```
bwoebi: For maximum efficiency please put the `instanceof \Generator` check first. `is_callable()` isn't totally inexpensive.

lib/functions.php
bwoebi: Varname is fine IMO. If it isn't a Generator, it will be one very quickly :-D

lib/functions.php
kelunik: We could also restrict it to `Closure`, since every other callable should be simply invoked and the `Generator` passed.

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
...
+ * @param \Generator|callable $generator A generator or callable the returns a generator to resolve as a coroutine
+ * @return \Amp\Promise
  */
-function resolve(\Generator $generator) {
+function resolve($generator) {
+    if (is_callable($generator)) {
+        $generator = call_user_func($generator);
+
+        if (!$generator instanceof \Generator) {
+            throw new \LogicException('Callable passed to resolve() did not return an instance of Generator');
```
bwoebi: Is that a LogicException or an InvalidArgument? Technically we have not provided a callable returning an instance of Generator, so should be invalid arg?

lib/functions.php
kelunik: `$generatorOrCallable` or `$generatorOrClosure` maybe?

lib/functions.php
kelunik: See also https://github.com/amphp/amp/pull/45#discussion_r62873782

lib/functions.php
bwoebi: technically, yes; but is there a reason to? If you come that way, we just as well may reject the whole PR?!

kelunik: > technically, yes; but is there a reason to?

`Closure` should be faster.

> If you come that way, we just as well may reject the whole PR?!

No, accepting `Closure` provides value, because it provides some sugar, especially for PHP 5, where you can't use the universal variable syntax. Accepting other values however, doesn't provide any additional value?

DaveRandom:
Exceptions updated, logic changes so that passing a generator (i.e. existing usage) only costs 1 extra `instanceof`

lib/functions.php
```diff
@@ -700,8 +700,8 @@ function timeout(Promise $promise, $msTimeout) {
  * Otherwise the promise's resolved value is returned.
  *
  * @param \Amp\Promise $promise The promise on which to wait
- * @throws \Exception if the promise fails
- * @return mixed Returns the eventual resolution result for the specified promise
+ * @return mixed if the promise fails
+ * @throws \Throwable
```
bwoebi: That comment is somewhat bogus now??

lib/functions.php
DaveRandom: Since the same var is re-used for the return value of invoking a callable, I personally think `$generator` is right

lib/functions.php
DaveRandom: wtf

DaveRandom:
I think I can now type `git commit -am temp && git rebase -i HEAD~2` faster than anyone else on earth.

bwoebi:
@DaveRandom you should use `git commit --amend -a`

DaveRandom:
I thought that didn't work sanely for commits that have already been pushed?

bwoebi:
Nah, it's totally fine

kelunik:
Just the same as `rebase`, it requires a force push.

lib/functions.php
```diff
@@ -752,9 +752,20 @@ function coroutine(callable $func) {
  * Upon resolution the Generator return value is used to succeed the promised result. If an
  * error occurs during coroutine resolution the returned promise fails.
  *
- * @param \Generator $generator The generator to resolve as a coroutine
+ * @param \Generator|callable $generator A generator or callable the returns a generator to resolve as a coroutine
```
kelunik: `s/the/that`

DaveRandom:
Ahh OK, I have that in my head from when I didn't understand git (so... like... right now, just like everyone else).

lib/functions.php
DaveRandom: F.
F.
S.

:-P

kelunik:
Where's the `IllegalArgumentException` now?

bwoebi:
@kelunik squashed into a single exception

kelunik:
@bwoebi The previous exceptions were better.

PR #53, opened 2016-12-09
Avoid fcall in Observer when possible

rdlowrey:


kelunik:
I just disabled comments by @coveralls, the status API is enough.

lib/Observer.php
```diff
@@ -122,7 +122,7 @@ public function next(): Promise {
 
         ++$this->position;
 
-        if (\array_key_exists($this->position, $this->values)) {
+        if (isset($this->values[$this->position]) || \array_key_exists($this->position, $this->values)) {
```
bwoebi: Please check whether this makes actually a benefit.

I believe calling next() with the value already existing is not happening that often in real world.

rdlowrey:
Not worth it. Nevermind.

PR #54, opened 2016-12-09
Make exception messages more generic

rdlowrey:
Since the exceptions don't have to be specifically tied to promise resolution I believe it makes more sense to make the default messages more generic.

### DIFF
--- a/lib/MultiReasonException.php
+++ b/lib/MultiReasonException.php
@@ -11,7 +11,7 @@ class MultiReasonException extends \Exception {
      * @param string|null $message
      */
     public function __construct(array $reasons, string $message = null) {
-        parent::__construct($message ?: "Too many promises were rejected");
+        parent::__construct($message ?: "Multiple errors encountered");
 
         $this->reasons = $reasons;
     }
--- a/lib/TimeoutException.php
+++ b/lib/TimeoutException.php
@@ -7,6 +7,6 @@ class TimeoutException extends \Exception {
      * @param string|null $message
      */
     public function __construct(string $message = null) {
-        parent::__construct($message ?: "Promise resolution timed out");
+        parent::__construct($message ?: "Operation timed out");
     }
 }


lib/Observer.php
```diff
@@ -122,7 +122,7 @@ public function next(): Promise {
 
         ++$this->position;
 
-        if (\array_key_exists($this->position, $this->values)) {
+        if (isset($this->values[$this->position]) || \array_key_exists($this->position, $this->values)) {
```
bwoebi: uh, wrong PR?

lib/TimeoutException.php
```diff
@@ -7,6 +7,6 @@ class TimeoutException extends \Exception {
      * @param string|null $message
      */
     public function __construct(string $message = null) {
-        parent::__construct($message ?: "Promise resolution timed out");
+        parent::__construct($message ?: "Operation timed out");
```
bwoebi: Uhm, Amp\TimeoutException is used exclusively for promise timeouts…!?

lib/Observer.php
rdlowrey: Yeah this was a derp. Ignore.

lib/TimeoutException.php
rdlowrey: Just because it doesn't have to be. It could be used for anything anyone wants it to be used for ...

rdlowrey: Otherwise it should be called `PromiseTimeoutException` IMO

bwoebi: In general you shall use library specific exceptions anyway, so no reason to.

bwoebi: Dunno, Amp is _the_ Promise library. no need to be more specific here. If you catch, you shall feel free to `use Amp\TimeoutException as PromiseTimeoutException;`

kelunik: `Operation timed out` is fine IMO if not even better.

staabm: Agree

lib/MultiReasonException.php
```diff
@@ -11,7 +11,7 @@ class MultiReasonException extends \Exception {
      * @param string|null $message
      */
     public function __construct(array $reasons, string $message = null) {
-        parent::__construct($message ?: "Too many promises were rejected");
+        parent::__construct($message ?: "Multiple errors encountered");
```
kelunik: I think we should stringify at least the exception names and messages into the message, otherwise they get lost.

PR #55, opened 2016-12-11
Rename Observer::next to Observer::hasNext

kelunik:
This ensures it matches the naming scheme of getCurrent and getResult.

### DIFF
--- a/lib/Observer.php
+++ b/lib/Observer.php
@@ -113,7 +113,7 @@ class Observer {
      *
      * @return \Interop\Async\Promise<bool>
      */
-    public function next(): Promise {
+    public function hasNext(): Promise {
         if (isset($this->deferreds[$this->position])) {
             $future = $this->deferreds[$this->position];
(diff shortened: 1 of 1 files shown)


rdlowrey:
👍 was thinking the same

bwoebi:
Shouldn't we instead rename getCurrent and getResult to result and current? (similarly to the Iterator interface)?

kelunik:
I thought about that as well, but I really don't like `->result()`.

joshdifabio:
Don't you think that `hasNext()` gives the impression that the method is side-effect-free? I wouldn't expect a method called `hasNext()` to change the state of the object given that there's no verb in the name.

kelunik:
Maybe we should just go with `advance`?

rdlowrey:
sign me up for `advance`

bwoebi:
Looks like 4 people like it … so lets go with it :-)

kelunik:
I'll push that one shortly.

kelunik:
See 1bbcada.

PR #57, opened 2016-12-23
Fix Promise::when invocation order

kelunik:


trowski:
Unfortunately this breaks the protection of double resolving a promise. Now a promise can be resolved again in the when callback.

test/WhenOrderTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+class WhenOrderTest extends \PHPUnit_Framework_TestCase {
+    public function testWhenOrder() {
+        $this->expectOutputString("1234");
+
+        $deferred = new Deferred;
+        $promise = $deferred->promise();
+
+        $promise->when(function () use ($promise) {
+            $promise->when(function () {
```
joshdifabio: This `when()` isn't invoked a synchronously with the current implementation. Is that your intention?

joshdifabio:
How about updating the test to check which `when()` calls are handled synchronously?

joshdifabio:
This is what I had in mind: https://github.com/amphp/amp/pull/58

test/WhenOrderTest.php
kelunik: I had not considered that. We have to define when a promise is resolved. I guess it should be _before_ the callbacks are invoked.

joshdifabio: I agree. I think it has to be before, otherwise we have fired the listeners for a promise when it's still in _pending_ state.

kelunik:
I think the current order is actually the right one and we should change the spec to account for that.

lib/Internal/Placeholder.php
```diff
@@ -3,7 +3,8 @@
 namespace Amp\Internal;
 
 use Amp\Failure;
-use Interop\Async\{ Loop, Promise };
+use Interop\Async\Loop;
```
trowski: IDE change this?

kelunik: Both, IDE changed it to

```
use Interop\Async\{
    Loop, Promise
};
```

I changed it then to avoid those stupid group uses.

trowski: That's annoying… probably isn't worth grouping them then unless there's 3 or more.

kelunik: It's never worth. Group use shouldn't be a thing. It's way harder to read and find the right entries.

joshdifabio:
Wait. Is it just me or is it impossible for the callbacks to be executed in the order they're added to the promise _and_ for `when()` calls on resolved promises to be fired synchronously? Looking at your test this seems to be the case.

kelunik:
@joshdifabio Right.

bwoebi:
@joshdifabio right. But that's an edge case though, I do not care too much; I think immediate resolution should take priority though (least surprise IMHO).

bwoebi:
Can we please close that PR and defer this to async-interop/promise-test?

kelunik:
@bwoebi No we can't. `async-interop/promise-test` only tests already resolved promises.

bwoebi:
@kelunik How do you think that? … no?!?

kelunik:
Nevermind.

PR #58, opened 2016-12-23
Add checks for synchronous when() invocation

joshdifabio:


test/WhenOrderTest.php
```diff
@@ -5,26 +5,44 @@
 use Amp\Deferred;
 
 class WhenOrderTest extends \PHPUnit_Framework_TestCase {
+    private $promise3ResolvedSynchronously;
```
kelunik: Why not just use references and `use`?

kelunik:
#57 tests for that now I think.

test/WhenOrderTest.php
joshdifabio: Yeah, that would work too.

PR #59, opened 2017-01-07
Make CallableMaker internal

kelunik:


bwoebi:
I see no advantage in marking this as internal.
The trait has a very well defined scope and task… It's unlikely to ever need an API change

kelunik:
@bwoebi If you follow that argumentation, `WhenQueue` shouldn't be internal as well. It doesn't really suit in the package. If we want to provide it as generic thing, we could make a separate package.

bwoebi:
WhenQueue though is a trait particularly designed for when()s, and we might later on decide that there is a more efficient way to manage that.
WhenQueue is rightfully internal here.

rdlowrey:
I would want to use this trait in most libraries built using amp; it's really useful. Moving it to internal would prevent that 😞

kelunik:
@bwoebi Once we require 7.1, this trait doesn't have any point anymore.

bwoebi:
@kelunik Once we require 7.1, _every_ library using Amp will also want to drop use of this trait - if they don't remove it by then, they would anyway end up with legacy artifacts.

kelunik:
@bwoebi The difference is that we need a new major version then.

bwoebi:
We anyway should use a new major version when increasing the base PHP version.

kelunik:
@bwoebi Why?

bwoebi:
Because we'll use functionality of that new language (a BC break) - which will be the whole point of bumping constraints?

kelunik:
@bwoebi Which is totally fine. Users not meeting the runtime requirement will just get the older version without new features.

bwoebi:
Good luck fucking things up. In theory that sounds fine … until some people develop e.g. on an a bit newer version of PHP and push an incompatible build or similar things.
I honestly wouldn't do that.

kelunik:
How is that related?

trowski:
Even when we require 7.1 we can leave this trait in place… it doesn't hurt anything and provides BC. We can remove it on a major version, whenever that may be.

PR #60, opened 2017-01-08
Add final where it's missing

kelunik:
We can always remove them later if someone really needs to extend one of the exception classes.

Maybe we should have an interface for Listener?

### DIFF
--- a/lib/InvalidYieldError.php
+++ b/lib/InvalidYieldError.php
@@ -2,7 +2,7 @@
 
 namespace Amp;
 
-class InvalidYieldError extends \Error {
+final class InvalidYieldError extends \Error {
     /**
      * @param \Generator $generator
      * @param string $prefix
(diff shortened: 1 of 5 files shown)


bwoebi:
a) Aerys is extending Listener (it's extending it with a Stream interface for getting the whole msg at once) [thus, just no.]
b) LazyPromise is fine with final
c) If you already acknowledge _if someone really needs to extend one of the exception classes_ the case, why do you even consider them as final? Please don't then.

kelunik:
> c) If you already acknowledge if someone really needs to extend one of the exception classes the case, why do you even consider them as final? Please don't then.

I only said that, because I assumed it's the reason they're not final. I don't see a real reason to extend these.

trowski:
Only one I have an issue with is `Listener` since that class is extended in a couple of Amp libs. I suppose we could make an interface and wrap it in those libs, but is that necessary?

bwoebi:
@kelunik I don't see a real reason for forbidding extension of the interfaces…

PR #71, opened 2017-03-12
AsyncInterop Merge

kelunik:
This merges the repositories `async-interop/event-loop`, `async-interop/event-loop-test`, `async-interop/promise-test` and `amphp/loop` into `amphp/amp`. It seems the `async-interop` project has failed at this point. It might be revived at a later point in time.

Full history is hereby retained and the necessary copyright notice merged into our `LICENSE` file.

This PR also adds auto-wrapping of created watchers, so they're automatically run as coroutines.


lib/Internal/Placeholder.php
```diff
@@ -5,6 +5,8 @@
 use Amp\Failure;
 use Amp\Loop;
 use Amp\Promise;
+use function Amp\adapt;
+use React\Promise\PromiseInterface as ReactPromise;
```
kelunik: If we follow the usual order, then this should be changed to 

```
use React\Promise\PromiseInterface as ReactPromise;
use function Amp\adapt;
```

trowski: Ok, no problem.

PR #76, opened 2017-03-13
Accept React promises in rethrow, wait, timeout, pipe, and capture

### LINKED ISSUE #75: Accept ReactPromise in rethrow, wait, timeout, pipe and capture
We support React promises in combinators, I think we should support them in those functions as well for consistency then?

trowski:
Fixes #75.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -91,9 +91,19 @@ function call(callable $functor, ...$args): Promise {
 /**
  * Registers a callback that will forward the failure reason to the Loop error handler if the promise fails.
  *
- * @param \Amp\Promise $promise
+ * @param \Amp\Promise|\React\Promise\PromiseInterface $promise
+ *
+ * @throws \TypeError If $promise is not an instance of \Amp\Promise or \React\Promise\PromiseInterface.
  */
-function rethrow(Promise $promise) {
+function rethrow($promise) {
+    if (!$promise instanceof Promise) {
+        if ($promise instanceof ReactPromise) {
+            $promise = adapt($promise);
+        } else {
+            throw new \TypeError("Must provide an instance of %s or %s", Promise::class, ReactPromise::class);
+        }
+    }
+
     $promise->when(function ($exception) {
         if ($exception) {
             throw $exception;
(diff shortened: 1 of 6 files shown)


lib/functions.php
```diff
@@ -104,13 +114,22 @@ function rethrow(Promise $promise) {
...
+ * @throws \TypeError If $promise is not an instance of \Amp\Promise or \React\Promise\PromiseInterface.
  * @throws \Throwable Promise failure reason.
  */
-function wait(Promise $promise) {
+function wait($promise) {
+    if (!$promise instanceof Promise) {
+        if ($promise instanceof ReactPromise) {
+            $promise = adapt($promise);
+        } else {
+            throw new \TypeError("Must provide an instance of %s or %s", Promise::class, ReactPromise::class);
```
kelunik: I think we should make a `_typeCheckFailed` function that accepts a value and an array of expected types. Helpful error messages should include the actual passed type.

test/TimeoutTest.php
```diff
@@ -8,6 +8,7 @@
 use Amp\Pause;
 use Amp\Success;
 use Amp\Promise;
+use function React\Promise\resolve;
```
kelunik: Should we alias those, too? All other imported React things are named `ReactPromise` or `RejectedReactPromise`.

lib/functions.php
trowski: How about a class extending `TypeError` with a constructor taking those arguments?

test/TimeoutTest.php
trowski: In a test… nah. I'd rather just use the fqn where it's called if you want to avoid ambiguity.

lib/functions.php
kelunik: That's fine with me, too.

test/TimeoutTest.php
kelunik: Fine with me either way, that's why it's a question. We don't have `resolve` anymore anyway.

lib/functions.php
kelunik: I think we can do that after the PR is merged.

bwoebi: I prefer the _typeCheckFailed; that way the class name will still be TypeError in the error log and not Amp\TypeError which may be confusing?

bwoebi:
Merged too fast … Don't the typehints need to accept ReactPromise[] too?

PR #86, opened 2017-03-14
Rename Promise::when to Promise::onResolve

### LINKED ISSUE #77: Changing "when" to "onResolve" or "onComplete"
It doesn't matter for normal promises, but I think it matters for more advanced use cases where the promise API is just one way of consumption. Looking at a list of functions, `onResolve` / `onComplete` are way more expressive than `when`.

kelunik:
Fixes #77.


bwoebi:
I said no. And I really, really dislike it. And I still say no.

PR #119, opened 2017-04-13
Review Changes

kelunik:
While reviewing v2 again, these are some of the remaining changes I would do. Further points not yet addressed:

- Kill Amp\Message
- Kill Amp\StreamIterator::drain()
- Add Amp\Promise::isResolved()?
- Find better names for Amp\wrap(), Amp\call()
- `git diff 2f7e9a0..a0b8de4 lib/Loop/UvDriver.php` → Cancel watchers instead of disabling them?


trowski:
- Kill Amp\Message: Yes, moving that to the stream lib.
- Kill Amp\StreamIterator::drain(): The method is needed for implementing the Message class. There's other ways, but they're much less efficient. I don't see any gain in not offering this method.
- Add Amp\Promise::isResolved(): I've considered this myself, it might eliminate the need for booleans that essentially just keep track of this, and is trivial to implement.
- Find better names for Amp\wrap(), Amp\call(): Very open to suggestions here… I have none.
- Cancel instead of disable: Could result in an invalid watcher being enabled again… but the original cause is an error anyway, so another potential error is probably fine.

Other changes LGTM. I did not realize using eval would stop IDE's complaining about multiple definitions, cool!

kelunik:
@trowski That's why we did it with `eval` already in version 1. It also stops IDEs from thinking `Deferred` implements `Promise`.

Regarding `drain`: If we want to keep it, we should probably add it to `Iterator`?

rybakit:
> That's why we did it with eval already in version 1. It also stops IDEs from thinking Deferred implements Promise.

This change confuses me:
1) Does `eval()` have support for OpCache?
2) What about debugging `eval`ed classes?
3) What if I still want to choose to what class declaration to jump?
4) Did you consider `@noinspection PhpUndefinedClassInspection` annotation?

bwoebi:
StreamIterator? … mhm, we anyway have Amp\Iterator, then StreamIterator is probably fine.

So: LGTM.

I dislike adding Amp\Promise::isResolved(). I do not know of real use cases, except inside tests. For tests it's trivial to add a helper function like
```php
function isPromiseResolved(Promise $p) {
    $resolved = false;
    $p->when(function() use (&$resolved) { $resolved = true; });
    return $resolved;
}
```

I kind of like \Amp\wrap() and \Amp\call(), I see no need for a rename here.

Regarding StreamIterator::drain(), I see no reason to object to Aaron.

Cancel vs. disable: I don't really care, if you get there, you already fucked up...

trowski:
@bwoebi `Promise::isResolved()` likely has little use, but `Deferred::isResolved()` and `Emitter::isResolved()` might be useful.

trowski:
@rybakit
1) Amp is for long running programs, OpCache matters little here.
2) What's there to debug? This is only for the production definitions of `Deferred` and `Emitter`, which are pulled from traits.
3) Not sure what you mean here… the definition is based on the value of `zend.assertions`.
4) `PhpUndefinedClassInspection` annotation ignores undefined classes. There is no way to ignore multiple definitions of which I'm aware.

bwoebi:
I'm not sure about these either. Typically you should get rid of Deferreds/Emitters once you have resolved them.
It feels for me like it would probably be a code smell to use isResolved(). If you really need to check whether it's been already resolved, one can simply _null_ it and check against that. But mostly there are better solutions except in highly specific cases.

trowski:
@bwoebi That's a good point. The few instances where `isResolved()` may be used, the deferred is nulled and likely then the preferred method as then a reference to the resolved value is not being kept unnecessarily. So then I agree, adding such a method has little value.

PR #121, opened 2017-04-15
Add missing namespace declarations

ck99:
Running in prod mode, the eval'd code needs a namespace declaration


trowski:
Apparently we need to try running in production mode more often. 😄

PR #123, opened 2017-04-23
Rename wrap and coroutine + various minor changes

kelunik:
This PR contains various minor changes, please review it with care. The major changes are `wrap` → `createRunnable` + `coroutine` → `createCallable` with the names borrowed from Java.

Further points not yet addressed:
 * `Coroutine::dispose` should probably not dispose the generator, but execute all finally blocks normally, just like a regular PHP function would. If another promise in a finally block fails, that error is the result of the promise instead, similar with returns.
* ~~`wait` should probably fail with an `Error` if `Loop::run` throws instead of letting that exception bubble up.~~ Addressed now.
* I'm not sure whether we should keep `pipe` and `capture`.


trowski:
- Not sure about `createRunnable` and `createCallable`… those names are going to need time to grow on me.
- Why the function instead of `UnionTypeError`? The same could be done for `InvalidYieldError`, so what's the difference?
- I used `UnionTypeError` in at least one place in another Amp lib, so if going with a function, maybe it should not be internal?
- I've been unsure of the behavior of `Coroutine::dispose` myself. Maybe it's better to change [these lines](https://github.com/amphp/amp/blob/master/lib/Coroutine.php#L123-L126) to simply `$this->generator = null` which should force `finally` blocks of the generator to be run.
- I agree on `wait`, adding the thrown exception from `Loop::run` as the previous exception.
- My initial reaction to eliminating `pipe` and `capture` was no… but really uses of those should probably be coroutines, so maybe it would be best to discourage `then`-style usage of promises.

kelunik:
`wait` throws an `Error` now in case `Loop::run` throws.

bwoebi:
`createCallable(callable): callable` - that signature makes absolutely no sense to me; you pass in a callable to create something it already is? `coroutine(callable): callable` makes much more sense in that context. At least one then knows that it's about creating callable coroutines. I do not see where it makes the name any clearer.

`createRunnable(callable)` … what's the difference between calling and running? if I call a function, it runs??

I think we should retain `coroutine()` and rename `wrap()` to `asyncCoroutine()`, because that's typically what it is - it creates a coroutine which will continue asynchronously (i.e. not along the flow of the calling coroutine).

I don't care about Amp\UnionTypeError or TypeError, in the end it's just an Error which anyway should not be caught.

The other changes are fine.

kelunik:
Renamed to `asyncCoroutine` and `coroutine`.

PR #126, opened 2017-04-27
Stream → Iterator Refactor

kelunik:
WIP.


examples/iterators/emitter.php
```diff
@@ -26,23 +23,15 @@
...
-
-            while (yield $listener->advance()) {
-                printf("Stream emitted %d\n", $listener->getCurrent());
-                yield new Pause(100); // Listener consumption takes 100 ms.
-            }
-
-            printf("Stream result %d\n", $listener->getResult());
-        };
-
-        yield new Coroutine($generator($stream));
```
kelunik: What was the reason for this to be wrapped in a coroutine before?

lib/Iterator.php
```diff
@@ -3,7 +3,7 @@
 namespace Amp;
 
 /**
- * Defines an asynchronous stream iterator that is designed to be used within a coroutine.
+ * Defines an asynchronous stream that is designed to be used within a coroutine.
```
kelunik: stream → iterator

examples/iterators/emitter.php
trowski: No idea honestly… probably some leftover from code being refactored from the async-interop loop. I think it was wrapped with `wait()` at one point.

test/FilterTest.php
```diff
@@ -3,109 +3,99 @@
...
-
-            $callback = function ($exception, $value) use (&$reason) {
-                $reason = $exception;
-            };
-
-            $stream->onResolve($callback);
+            try {
+                while (yield $iterator->advance()) {
+                    $iterator->getCurrent();
+                }
```
kelunik: There should probably be a `fail` here?

test/FilterTest.php
```diff
@@ -3,109 +3,99 @@
...
 
             $emitter->fail($exception);
 
-            $callback = function ($exception, $value) use (&$reason) {
-                $reason = $exception;
-            };
+            try {
+                while (yield $iterator->advance()) {
+                    $iterator->getCurrent();
+                }
```
kelunik: Same here, probably a `fail`.

test/FilterTest.php
trowski: Yep, there's several other places with the same pattern. A caveat is that `catch` will catch the test failure, so maybe I need a specific exception class too.

lib/Emitter.php
```diff
@@ -25,32 +26,32 @@
...
                 );
             }
 
             /**
-             * @return \Amp\Stream
+             * @return \Amp\Iterator
              */
-            public function stream(): Stream {
-                return $this->stream;
+            public function getIterator(): Iterator {
```
bwoebi: This should probably be called `iterate()`

lib/Emitter.php
```diff
@@ -61,16 +62,14 @@ public function emit($value): Promise {
...
 
             /**
-             * Resolves the stream with the given value.
-             *
-             * @param mixed $value
+             * Completes the iterator.
              */
-            public function resolve($value = null) {
-                ($this->resolve)($value);
+            public function complete() {
```
bwoebi: `complete()` or `end()`?

lib/Emitter.php
kelunik: We had a discussion in chat about it.

bwoebi: Then I must have missed said discussion. Can you link me to it?

lib/Emitter.php
kelunik: I like `complete()`, less risk of method name collision with other things that _might_ also implement `Iterator`.

lib/Emitter.php
kelunik: Sure. http://chat.stackoverflow.com/transcript/11?m=36837458#36837458

bwoebi: Well, I don't buy your argument - I don't really see how this will make it easier to distinguish that.

Additionally, technically, `iterate()` could also each time return an other object. Promise and Iterator have both their state managed by another entity.

I'll go with my OCD and prefer symmetry here.

kelunik: `iterate` implies it's from the beginning for me.

bwoebi: The same way you could argue that `getIterator` shall return _an_ Iterator from the beginning.
At least that's what `IteratorAggregate::getIterator()` does - it normally returns you a new iterator instance.
Retaining that symmetry there may let you have false expectations.

trowski: I would also prefer `Emitter::iterate()` here, partly for symmetry but also because @bwoebi makes a good point with `IteratorAggregate` returning a new instance, so too would be the assumption here with `getIterator()`.

I think docs will just need to be clear here regardless of which name is chosen, so in that case symmetry wins out IMO.

kelunik:
Rebased and merged manually.

PR #127, opened 2017-04-28
Rename Pause to Delayed

kelunik:
Pause doesn't cover the delayed value use case.

### DIFF
--- a/examples/streams/backpressure.php
+++ b/examples/streams/backpressure.php
@@ -4,9 +4,9 @@
 require __DIR__ . '/../../vendor/autoload.php';
 
 use Amp\Coroutine;
+use Amp\Delayed;
 use Amp\Emitter;
 use Amp\Loop;
-use Amp\Pause;
 
 Loop::run(function () {
     try {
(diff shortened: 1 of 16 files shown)


kelunik:
@bwoebi @trowski Opinions?

bwoebi:
Not really. I consider it a pretty unnecessary rename: `delay continuing execution for X seconds, then return a value` vs `pause for X seconds, then return a value`.
I see no benefit of renaming, except a little bit potential confusion.

kelunik:
@bwoebi It doesn't really pause execution, it's a promise like everything else. And if it's a pause, it shouldn't have a value IMO.

bwoebi:
It's a pause within the synchronous context of a coroutine/promise.
Not sure how Pause/Delay are related to the value?

kelunik:
[That's how the value is related.](https://github.com/amphp/amp/blob/cc32620f5c256c35cd86c27b560476f5ef2b2303/lib/Pause.php#L16). Note that the suggestion is `Delayed`, not `Delay`. `Delay` would be no difference to `Pause`.

trowski:
`Delayed` makes more sense outside of a coroutine and has a nice duality with `Loop::delay()`. `Pause` always felt weird… pause what?

bwoebi:
I was reading `Delay` in my mind … well, Delayed is fine.

PR #133, opened 2017-05-19
Allow setting the loop driver via the env

kelunik:
See http://chat.stackoverflow.com/transcript/message/37192009#37192009.


bwoebi:
I'm going to merge that for now, we always can add a shortcut option later.

PR #136, opened 2017-05-21
Implement IteratorTransform to allow fluent transformation

### LINKED ISSUE #109: Enable fluent interface for streams
The stream functions suffer from the same problems as PHP's array functions; they're awkward to chain.

```php
$output = filter(map(filter($input, $filterFn), $mapFn), $otherFilterFn);
```

This would be preferable imo:

```php
$output = $input
    ->apply(filter($filterFn))
    ->apply(map($mapFn))
    ->apply(filter($otherFilterFn));
```

Edit: Inspired by apache beam.

kelunik:
This allows chaining transformations. Closes #109.

I'm not really sure we need something like that.

/cc @joshdifabio

### DIFF
--- a/lib/IteratorTransform.php
+++ b/lib/IteratorTransform.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Amp;
+
+/**
+ * Allows consecutive transformation of an iterator.
+ */
+class IteratorTransform implements Iterator {
+    private $applied = false;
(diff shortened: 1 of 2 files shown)


kelunik:
Closing. Can be revisited later. Currently we do not see the need for this.

http://chat.stackoverflow.com/transcript/message/37230418#37230418

PR #137, opened 2017-05-23
Require minimum of 1ms per timer watcher, fixes #131

### LINKED ISSUE #131: Inconsistent repeat watchers with interval = 0
```php
$detector = new BlockDetector(..., $threshold, $interval);
$detector->start();

Loop::run(function () use ($detector) {
    Loop::repeat(0, function () {
        usleep(100 * 1000);
    });

    Loop::delay(300, function () {
        Loop::stop();
   });
});
```

Currently I have a bug that doesn't call the detector (migrated from the interop loop, worked before) with UV. With the `NativeDriver` it completely blocks and never stops.

kelunik:
UV and EV have issues with the testExecutionOrderGuarantees test now.

@bwoebi @trowski  Could anyone look into these? The `usleep(1000)` fixes it for Native and Event.

### DIFF
--- a/lib/Loop/Driver.php
+++ b/lib/Loop/Driver.php
@@ -193,7 +193,7 @@ abstract class Driver {
      * The created watcher MUST immediately be marked as enabled, but only be activated (i.e. callback can be called)
      * right before the next tick. Callbacks of watchers MUST NOT be called in the tick they were enabled.
      *
-     * @param int   $delay The amount of time, in milliseconds, to delay the execution for.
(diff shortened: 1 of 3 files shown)


bwoebi:
Uh, do we also need minimum 1ms for delay watchers? Thought only for repeat watchers?

kelunik:
Fixed via https://github.com/amphp/amp/commit/38ef4d9ffac77e4d353f3e15bacd0d7db5d6ec29.

PR #146, opened 2017-07-04
Support yielding promise => cancellation-token in coroutines

trowski:
This PR uses the `key => value` generator syntax to support yielding a cancellation token with a promise as a convenient way of providing an escape to yielding promises that slowly or never resolve.

### DIFF
--- a/lib/Coroutine.php
+++ b/lib/Coroutine.php
@@ -119,7 +119,21 @@ final class Coroutine implements Promise {
                 $promise = $this->generator->key();
 
                 if (!$promise instanceof Promise) {
-                    throw new \Error("Must yield a promise as the key when yielding a cancellation token");
+                    if (\is_array($promise)) {
+                        $promise = Promise\all($promise);
+                    } elseif ($promise instanceof ReactPromise) {
+                        $promise = Promise\adapt($promise);
+                    } else {
+                        return new Failure(new InvalidYieldError(
+                            $this->generator,
+                            \sprintf(
+                                "Key must be an instance of %s or %s or array of such instances when yielding an instance of %s",
+                                Promise::class,
+                                ReactPromise::class,
+                                CancellationToken::class
+                            )
+                        ));
+                    }
                 }
 
                 return Promise\cancellable($promise, $yielded);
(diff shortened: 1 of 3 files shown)


lib/Coroutine.php
```diff
@@ -115,6 +115,16 @@ private function transform($yielded): Promise {
                 return Promise\adapt($yielded);
             }
 
+            if ($yielded instanceof CancellationToken) {
+                $promise = $this->generator->key();
+
+                if (!$promise instanceof Promise) {
```
kelunik: This should also support `ReactPromise` here.

lib/functions.php
```diff
@@ -238,6 +239,46 @@ function adapt($promise): Promise {
...
+            }
+        }
+
+        $deferred = new Deferred;
+        $id = $token->subscribe([$deferred, "fail"]);
+
+        $promise->onResolve(function ($exception, $value) use ($id, $token, $deferred) {
+            if ($token->isRequested()) {
+                return; // Deferred has already been resolved.
+            }
```
kelunik: The code doesn't do here what the comment says.

trowski: If cancellation has already been requested, then the deferred should have already been failed in the subscribe callback. Am I missing something?

lib/Coroutine.php
trowski: Should probably support an array too, I'll refactor.

lib/functions.php
kelunik: I guess you're right, it's just a bit weird. :)

trowski:
Thoughts on this? Is this providing a convenient syntax for a common use-case or hiding too much?
```php
yield $socket->enableCrypto() => new TimeoutCancellationToken(100);
// vs.
yield \Amp\Promise\cancellable($socket->enableCrypto(), new TimeoutCancellationToken(100));
```

joshdifabio:
The short syntax is a bit weird imo. Sorry to be negative.

trowski:
@joshdifabio No reason to be sorry. @kelunik and I just came up with this idea and we're looking for feedback. The function `Amp\Promise\cancellable()` will definitely be added (not necessarily with that name), but I'm not sure about the coroutine syntax yet.

trowski:
Closing this as it seemed like an interesting approach, but overall it's unclear what's happening. I'll consider opening another PR for just the `cancellable` function.

PR #162, opened 2017-09-16
Ignore only warnings for interrupted syscalls in NativeDriver

### LINKED ISSUE #160: stream_select error suppression
Currently we suppress errors for `stream_select`, but that might also hide useful error messages like the one shown in https://github.com/ratchetphp/Ratchet/issues/300:

```
PHP Warning:  stream_select(): You MUST recompile PHP with a larger value of FD_SETSIZE.
It is set to 1024, but you have descriptors numbered at least as high as 1123.
 --enable-fd-setsize=2048 is recommended, but you may want to set it
to equal the maximum number of open files supported by your system,
in order to avoid seeing this error again at a later date.
```

Dunno whether we can do something about that without a high performance impact.

kelunik:
Fixes #160.

### DIFF
--- a/lib/Loop/NativeDriver.php
+++ b/lib/Loop/NativeDriver.php
@@ -144,7 +144,13 @@ class NativeDriver extends Driver {
 
             // Error reporting suppressed since stream_select() emits an E_WARNING if it is interrupted by a signal.
             if (!@\stream_select($read, $write, $except, $seconds, $microseconds)) {
-                return;
+                $error = error_get_last();
+
+                if (\strpos($error["message"], "Interrupted system call") !== false) {
+                    return;
+                }
+
+                $this->error(new \Exception("stream_select() failed: " . $error["message"]));
             }
 
             foreach ($read as $stream) {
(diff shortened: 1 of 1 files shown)


lib/Loop/NativeDriver.php
```diff
@@ -144,7 +144,13 @@ private function selectStreams(array $read, array $write, $timeout) {
...
             // Error reporting suppressed since stream_select() emits an E_WARNING if it is interrupted by a signal.
             if (!@\stream_select($read, $write, $except, $seconds, $microseconds)) {
-                return;
+                $error = error_get_last();
+
+                if (\strpos($error["message"], "Interrupted system call") !== false) {
+                    return;
+                }
+
+                $this->error(new \Exception("stream_select() failed: " . $error));
```
jsor: Array to string conversion for `$error` here.

kelunik: Good catch!

kelunik:
Manually merged via a2477575355d1871779974c9e5902b70955071fb.

PR #164, opened 2017-09-18
Change getProtocol => getProtocolVersion

sedatsevgili:
In current("^3.0") amphp/artax version, there is no getProtocol method, instead of getProtocolVersion method exists in artax response. If you try, you will see kind of this error: 

"Error: Call to undefined method class@anonymous::getProtocol() in foobar.php on line xxx"


kelunik:
Thanks!

sedatsevgili:
@kelunik you're welcome!

PR #166, opened 2017-09-20
added example how to treat exceptions

umbri:
I think this must be somewhere in docs not in examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+### 2.0.2
+
+ - Fixed warnings and timers in `EventDriver`.
+ - Does no longer hide warnings from `stream_select`.
+
+### 2.0.1
+
+ - Fixed an issue where the loop blocks even though all watchers are unreferenced.
+
 2.0.0
 -----
 


examples/exceptions/exception.php
```diff
@@ -0,0 +1,28 @@
...
+
+require_once __DIR__ . "/../../vendor/autoload.php";
+
+function asyncOperation() {
+    $def = new Deferred();
+
+    \Amp\Loop::delay(1000, function () use ($def) {
+        $def->fail(new Exception("force exception")); //this will be catch-ed
+
+//        throw new Exception("force exception"); //this will be NOT catch-ed
```
kelunik: I think there should be a second async function that throws inside the `delay` callback and a second `Loop::run`.

kelunik:
@umbri Thanks! I added a comment. Maybe you also want to fix your email, so the commit is properly linked with your GitHub account.

examples/exceptions/exception.php
umbri: @kelunik OK will add 2 example catch-ed & uncatch-ed and PR

umbri:
@kelunik Thank you! I will fix it

examples/exceptions/uncatched.php
```diff
@@ -0,0 +1,26 @@
...
+
+Loop::run(function () {
+    try {
+        $res = yield asyncOperation();
+
+        echo "result -> " . $res . PHP_EOL;
+    } catch (Throwable $e) {
+        echo "catch -> " . $e->getMessage() . PHP_EOL;
+    }
+});
```
kelunik: I'd add a second `try { ... } catch` around the `Loop::run` to show that exceptions thrown in loop callbacks just bubble up.

PR #167, opened 2017-09-20
added exception examples

umbri:


umbri:
@kelunik Sorry, I pushed 1 commit with my private email and have to remove it, so why I create another PR

kelunik:
@umbri Then just force-push to the previous branch, so your old commit goes away?

umbri:
@kelunik Yes, goes away. I am sorry about your comments on that PR, I split in 2 examples, as you have suggested in previous PR

examples/exceptions/uncatched.php
```diff
@@ -0,0 +1,26 @@
...
+
+Loop::run(function () {
+    try {
+        $res = yield asyncOperation();
+
+        echo "result -> " . $res . PHP_EOL;
+    } catch (Throwable $e) {
+        echo "catch -> " . $e->getMessage() . PHP_EOL;
+    }
+});
```
kelunik: As mentioned in the other PR, please add a `try { ... } catch {  ... }` around the `Loop::run()`.

examples/exceptions/uncatched.php
```diff
@@ -0,0 +1,26 @@
...
+<?php
+use Amp\Deferred;
+use Amp\Loop;
+
+require_once __DIR__ . "/../../vendor/autoload.php";
+
+function asyncOperation() {
+    $def = new Deferred();
+
+    \Amp\Loop::delay(1000, function () use ($def) {
```
kelunik: `\Amp\Loop` is already imported, so you can just write `Loop::delay` here.

examples/exceptions/uncatched.php
kelunik: Filenames should be `caught.php` / `uncaught.php`. Or maybe `try-catch.php` and `throwing-loop-callables.php`.

umbri:
@kelunik omg, I didn't know this is possible:
```
try {
    Loop::run(function () {
        //some more code here 
    });
} catch (Throwable $loopException) {
    echo "loopException -> " . $loopException->getMessage() . PHP_EOL;
}

```

kelunik:
Manually merged via bb34b9fb2d599375b1e34b10d7632ffd50d1fecb and improved via f2a662534e5e4f6736c841c11c11a2ddfaa85a5f.

PR #172, opened 2017-10-10
Preserve order of Amp\Promise\all

brstgt:
…mp\Promise\all()

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -284,7 +284,7 @@ namespace Amp\Promise {
                 throw createTypeError([Promise::class, ReactPromise::class], $promise);
             }
 
-            $promise->onResolve(function ($exception, $value) use (&$deferred, &$values, &$pending, $key) {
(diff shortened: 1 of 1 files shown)


lib/functions.php
```diff
@@ -284,7 +284,7 @@ function all(array $promises): Promise {
                 throw createTypeError([Promise::class, ReactPromise::class], $promise);
             }
 
-            $promise->onResolve(function ($exception, $value) use (&$deferred, &$values, &$pending, $key) {
+            $promise->onResolve(function ($exception, $value) use (&$deferred, &$values, &$pending, $key, &$promises) {
```
kelunik: There's no need to make this `use()` a reference.

kelunik:
Please add an entry to `CHANGELOG.md`.

bwoebi:
I've closed this PR in favor of a simpler commit, but thanks for the suggestion :-)

PR #173, opened 2017-10-16
Defined the @class docblock property

aaron-hall-7:
Defining the @class property in the initial doc block should allow this to be picked up in PHPStorm's auto-complete.

Tested by setting $def = new Def... and allowing PHPStorm to suggest Amp\Deferred

Tested by setting $emit = new Emit... and allowing PHPStorm to suggest Amp\Emitter

### DIFF
--- a/lib/Deferred.php
+++ b/lib/Deferred.php
@@ -2,7 +2,10 @@
 
 namespace Amp;
 
-// @codeCoverageIgnoreStart
+/**
+ * @codeCoverageIgnoreStart
+ * @class Deferred
+ */
 try {
     if (!@\assert(false)) {
         development: // PHP 7 development (zend.assertions=1)
--- a/lib/Emitter.php
+++ b/lib/Emitter.php
@@ -2,7 +2,10 @@
 
 namespace Amp;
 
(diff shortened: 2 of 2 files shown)


kelunik:
Unfortunately, it doesn't work for me. I installed it as `dev-phpstorm-deferred-autocomplete`, but there's no such suggestion. :-(

The strange thing is that if I write it out and hit `Alt + Enter`, the import is suggested correctly even without this patch.

aaron-hall-7:
Yeah it's working for me on 2017.2.4.  If you let me know what version you have, I can try and diagnose it for a specific version.  Otherwise, maybe I have a plugin that parses my docblocks for that information

screen shot 2017-10-19 at 9 25 58 pm
screen shot 2017-10-19 at 9 27 15 pm
screen shot 2017-10-19 at 9 27 37 pm

kelunik:
Same version here. I can confirm that it works when I checkout your branch locally. But it doesn't work when `amphp/amp` is pulled as dependency. Maybe that's just a PhpStorm bug?

kelunik:
Seems like it also works even without the `@class` if I open the project directly.

PR #176, opened 2017-10-31
@Class docblock

yuganshchokra:
Defining the @Class property in the initial doc block should allow this to be picked up in PHPStorm's auto-complete.

Tested by setting $def = new Def... and allowing PHPStorm to suggest Amp\Deferred

Tested by setting $emit = new Emit... and allowing PHPStorm to suggest Amp\Emitter


kelunik:
Please stop spamming.

PR #186, opened 2017-11-28
Automatically add \ for buitins. Fixes #185

### LINKED ISSUE #185: Php-cs-fixer to auto add \ for php builtins
Would be cool to have a php-cs-fixer which auto adds \ for php bultin calls, so we dont need to do the tedious task

As this is usefull for hot path e.g. bootstrapping logic in a lot of libs/components/frameworks this might even lbe usefull outside amp context

staabm:


kelunik:
Please also fix the code style then.

staabm:
here we go

staabm:
hmm not sure why my cs-fixer did not fix what travis is talking about... need to check that

staabm:
updating my fixer did the trick it seems

kelunik:
Thanks!

PR #187, opened 2017-11-30
Add strict_types + clean type-covered docblocks

TomasVotruba:
Brought to you by [EasyCodingStandard](https://github.com/Symplify/EasyCodingStandard)

### DIFF
--- a/lib/CallableMaker.php
+++ b/lib/CallableMaker.php
@@ -16,7 +16,6 @@ if (\PHP_VERSION_ID < 70100) {
          * publicly invokable callback.
          *
          * @param string $method Instance method name.
-         *
          */
         private function callableFromInstanceMethod(string $method): callable {
             if (!isset(self::$__reflectionMethods[$method])) {
@@ -34,7 +33,6 @@ if (\PHP_VERSION_ID < 70100) {
          * publicly invokable callback.
          *
          * @param string $method Static method name.
-         *
          */
         private static function callableFromStaticMethod(string $method): callable {
             if (!isset(self::$__reflectionMethods[$method])) {
--- a/lib/CallableMaker.php
+++ b/lib/CallableMaker.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 
 namespace Amp;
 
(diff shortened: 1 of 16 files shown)


PeeHaa:
Why did you remove the `void` return types?

PeeHaa:
Also I have a feeling you are in for a fight when trying to add strict types.

kelunik:
Moving to strict types would be a BC break and we explicitly decided not to go with strict types.

TomasVotruba:
> Also I have a feeling you are in for a fight when trying to add strict types.

What do you mean by that?

> kelunik

This would be targetting next major release of course.

kelunik:
The next major release is not planned, yet, and likely to be released in the far future. It doesn't make sense to have code-style PRs for a future major release, yet.

Which kind of additional feedback would you like to have?

PeeHaa:
Void type is only available as of 7.1... This library support 7.

Even if this library would have supported 7.1+ you just removed the doc block return type and didn't actually add the return type. So basically you just removed information.

TomasVotruba:
Usually people add these when bumping to PHP 7 to decrease upgrading for end users. I thought you'd be open to this. It looked like it was forgotten.

Well, the doc cleanup.

PeeHaa:
> > Also I have a feeling you are in for a fight when trying to add strict types.

> What do you mean by that?

Because I have had some discussion about this some while ago with the maintainers and their stance has been: nope, don't we don't do strict types.

kelunik:
Usually the return documentation blocks should have a description added instead of being removed. And strict types have been discussed and it was explicitly decided to avoid using them due to callbacks.

TomasVotruba:
@PeeHaa Ah, I didn't know that. Could you share a link?

TomasVotruba:
> @kelunik  Usually the return documentation blocks should have a description added instead of being removed.

Could explain why?

> decided to avoid using them due to callbacks.

I have no idea what this means. Could you explain in code please?

PeeHaa:
> decided to avoid using them due to callbacks.

Because the strict types declaration of callback are not defined where you think they are.

TomasVotruba:
> Because the strict types declaration of callback are not defined where you think they are.

Code would be more helpful, I don't use callback much so I don't know what you mean.

PR #188, opened 2017-12-02
Define AMP_DEBUG through an environment variable

trowski:
Easier to set up test environments, particularly Travis.

### DIFF
--- a/lib/Internal/Placeholder.php
+++ b/lib/Internal/Placeholder.php
@@ -94,7 +94,8 @@ trait Placeholder {
         }
 
         \assert((function () {
-            if (\getenv("AMP_DEBUG") || (\defined("AMP_DEBUG") && \AMP_DEBUG)) {
+            $env = \getenv("AMP_DEBUG");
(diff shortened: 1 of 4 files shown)


lib/Internal/Placeholder.php
```diff
@@ -84,15 +84,18 @@ private function resolve($value = null) {
...
+                // @codeCoverageIgnoreEnd
             }
 
             throw new \Error($message);
         }
 
         \assert((function () {
-            if (\defined("AMP_DEBUG") && \AMP_DEBUG) {
-                $trace = \debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+            if (\getenv("AMP_DEBUG") || (\defined("AMP_DEBUG") && \AMP_DEBUG)) {
```
staabm: Should the env also be checked for a truethy value?

trowski: It is, but I think I should provide a check against `"false"` if that's what you mean.

Issue #157, opened 2017-08-30
Incorrect assumption assertions are always enabled


far-blue:
The amp codebase behaves differently when assertions are enabled in ways beyond just using assert() - such as changes to class inheritance. The codebase allows invalid behaviour when assertions are disabled and there is an implicit assumption that developers will always be developing in a php environment where assertions are enabled. If developers don't enable assertions then invalid behaviour is allowed in the code which might then break in a different environment.

The assumption that all developers will have assertions enabled is false because the default installs of php in all operating system environments is production mode (with assertions disabled). The resulting developer experience is pretty poor and is likely to cause more issues for developers new to Amp than benefits.

Because the amp codebase effectively requires assertions to be enabled in order to safely use it for development, I suggest a warning be output to stdout when the run loop is enabled when assertions are disabled, unless the warning is explicitly disabled in code. There are already similar warnings in Aerys regarding debug mode and assertions.

bwoebi:
That may indeed be a problem.

With warning to stdout you probably mean `trigger_error()` with `E_USER_WARNING`?

I'm not sure whether that's the right solution … Perhaps someone else has good opinions on this?

kelunik:
I guess we should just disable that optimization by default and use `defined("AMP_OPTIMIZE_DEFERRED") && AMP_OPTIMIZE_DEFERRED` instead of assertions.

bwoebi:
I'd be totally generic by default and add an constant `AMP_OPTIMIZATIONS`. Then we can have `defined("AMP_OPTIMIZE_DEFERRED") ? AMP_OPTIMIZE_DEFERRED : (defined("AMP_OPTIMIZATIONS") && AMP_OPTIMIZATIONS)`. To have a super-switch turning all optimizations on/off, while still allowing particular optimizations to be enabled or disabled individually.

That super-switch constant shall then also be used in other amp projects.

We even might require that constant to be defined in applications or trigger a warning then - not sure about that.

kelunik:
@trowski What's your opinion on that?

trowski:
I like the idea that @bwoebi proposed above, particularly about issuing a warning if `AMP_OPTIMIZATIONS` is not defined so that users are aware that performance can be improved in production by defining the constant. I'll put together a PR in the coming days.

kelunik:
@trowski I don't like to throw a warning by default.

bwoebi:
@kelunik I agree, we should rather throw an exception there.

kelunik:
[...]

PR #189, opened 2017-12-02
Use env or const instead of assertions to enable/disable optimizations

trowski:
Addresses #157.

Instead of using assertions to toggle Amp's `Deferred`/`Emitter` optimizations, this PR uses either an environment variable or const `AMP_OPTIMIZATIONS`. If neither of these are defined, an `E_USER_NOTICE` is issued.

I think this is a better approach to enabling optimizations, as it is more obvious to the user and does not depend the setting for assertions, which commonly are configured incorrectly for the environment, tending to default to off.

I put the code to issue the notice in the `Promise` class definition file. This was a rather arbitrary place (only chosen because the optimizations relate to promises) and if someone has a better suggestion please let me know.

### DIFF
--- a/lib/Deferred.php
+++ b/lib/Deferred.php
@@ -9,7 +9,7 @@ if ((function () {
     }
     return $env === "0" || $env === "false";
 })()) {
-    development: // PHP 7 development (zend.assertions=1)
+    development: // PHP 7 development
     /**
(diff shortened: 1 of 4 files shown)


trowski:
Closed in favor of #190.

PR #190, opened 2017-12-03
Drop multiple definitions of Deferred and Emitter

trowski:
Alternative to #189, simply dropping the optimization when assertions are enabled.

`Deferred` and `Emitter` return objects with public methods that allow resolution/emitting/etc., but IDEs (at least not PhpStorm) will not hint those methods. I think the potential for abuse is slim, it's fairly clear that those methods are not suppose to be used. Such code would likely be fragile or throw when the promise is resolved again.

As @kelunik mentioned, the use of `Deferred` is much lower with Amp v2, particularly in user-code and high-level Amp libs which mostly use coroutines. Between this and the performance gains of PHP 7, perhaps this performance hack isn't necessary anymore.

### DIFF
--- a/lib/Deferred.php
+++ b/lib/Deferred.php
@@ -2,73 +2,46 @@
 
 namespace Amp;
 
-// @codeCoverageIgnoreStart
-try {
-    if (!@\assert(false)) {
-        development: // PHP 7 development (zend.assertions=1)
-        /**
-         * Deferred is a container for a promise that is resolved using the resolve() and fail() methods of this object.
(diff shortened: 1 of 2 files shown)


trowski:
Apparently fixing a typo in the branch name closes the pull request.

kelunik:
@bwoebi @rdlowrey Please review and respond whether you're fine with this change.

rdlowrey:
👍

trowski:
Merged via https://github.com/amphp/amp/commit/be3ae73a451e933d584ce9d9906776a11450c87a.

PR #192, opened 2017-12-19
Fix segfault with ev on shutdown, fixes amphp/parallel#3

kelunik:


lib/Loop/EvDriver.php
```diff
@@ -125,6 +125,8 @@ public function __destruct() {
         foreach ($this->events as $event) {
             $event->stop();
         }
+
+        $this->events = [];
```
trowski: Let's add a comment here as to why this is here, because it otherwise seems pointless.

PR #197, opened 2017-12-31
Hide resolution methods

trowski:
This modifies `Deferred` and `Emitter` to create another object without resolution methods that can be returned from `promise()` and `iterate()` respectively. This does have a slight negative performance impact, but would completely prevent users from shooting themselves in the foot. As noted in #190, most Amp code uses coroutines now, so I think actual performance impact is negligible.

### DIFF
--- a/lib/Deferred.php
+++ b/lib/Deferred.php
@@ -8,14 +8,28 @@ namespace Amp;
  * API, but used internally to create and resolve a promise.
  */
 final class Deferred {
-    /** @var \Amp\Promise */
+    /** @var object Has public resolve and fail methods. */
+    private $resolver;
+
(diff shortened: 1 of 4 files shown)


trowski:
Note if #198 is merged, this will need to be rebased first to drop the changes in `Emitter`.

kelunik:
Should we merge this?

lib/Deferred.php
```diff
@@ -8,16 +8,21 @@
  * API, but used internally to create and resolve a promise.
  */
 final class Deferred {
-    /** @var \Amp\Promise */
+    /** @var object Has public resolve and fail methods. */
```
staabm: It still implements Promise therefore could be typed like that

Issue #168, opened 2017-09-26
Iterators and garbage collection


kelunik:
I'm currently thinking about https://github.com/amphp/aerys/issues/109 and iterators. What happens if one responder returns a response with an `IteratorStream` and keeps an `Emitter` to streaming some events. If a middleware later replaces that stream without consuming it, the emitter will be stuck waiting for the backpressure to resolve, no?

An iterator not being consumed any further might also happen in other cases, which leaves the emitter without any notice to clean up, no?

trowski:
@kelunik It seems we need to take more care to have destructors on objects relying on iterators (such as `IteratorStream`) to consume the remaining values from the iterator to avoid the problem you described. I recently added such a destructor in the postgres lib to consume remaining tuples in unbuffered result sets: https://github.com/amphp/postgres/blob/master/lib/PqUnbufferedResult.php#L41-L45

Another option may be adding a `dispose()` method to iterators to indicate the consume is no longer interested in the values emitted. Back-pressure promises are currently resolved with null. However, we could resolve with a bool to indicate if the consumer has disposed – true if more values are desired, false if the iterator has been disposed.

kelunik:
Iterators have been removed in v3 due to these issues and will be replaced with pipelines: https://github.com/amphp/pipeline

PR #198, opened 2017-12-31
Relieve back-pressure on iterator when destoryed

trowski:
This provides a mechanism that relieves all back-pressure and fails future emits with an instance of `Amp\DisposedException` when the consumer of the iterator object destroys the object returned from `Producer::iterate()`. Currently the back-pressure promise never resolved, potentially leading to memory leaks or hanging.

### DIFF
--- a/lib/Emitter.php
+++ b/lib/Emitter.php
@@ -13,7 +13,7 @@ final class Emitter {
     private $iterator;
 
     public function __construct() {
-        $this->iterator = new class implements Iterator {
+        $this->iterator = new class {
             use Internal\Producer {
                 emit as public;
                 complete as public;
@@ -26,7 +26,7 @@ final class Emitter {
      * @return \Amp\Promise
      */
     public function iterate(): Iterator {
-        return $this->iterator;
+        return $this->iterator->iterate();
     }
 
     /**
--- a/lib/Internal/Producer.php
+++ b/lib/Internal/Producer.php
@@ -2,8 +2,11 @@
 
 namespace Amp\Internal;
 
+use Amp\CallableMaker;
 use Amp\Deferred;
+use Amp\DisposedException;
 use Amp\Failure;
+use Amp\Iterator;
 use Amp\Promise;
 use Amp\Success;
 use React\Promise\PromiseInterface as ReactPromise;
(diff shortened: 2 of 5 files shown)


lib/Internal/Producer.php
```diff
@@ -26,60 +26,115 @@
...
+                $this->complete = &$complete;
+            }
 
-        if (isset($this->backPressure[$this->position])) {
-            $future = $this->backPressure[$this->position];
-            unset($this->values[$this->position], $this->backPressure[$this->position]);
-            $future->resolve();
-        }
+            public function __destruct() {
+                if (!empty($this->backPressure)) {
```
kelunik: This should just be `if ($this->backPressure)`

lib/Internal/Producer.php
```diff
@@ -26,60 +26,115 @@
...
+                    throw new \Error("Promise returned from advance() must resolve before calling this method");
+                }
+
+                return $this->values[$this->position];
+            }
+        };
     }
 
     /**
-     * Emits a value from the iterator. The returned promise is resolved with the emitted value once all listeners
```
kelunik: once it has been consumed.

test/ProducerTraitTest.php
```diff
@@ -201,16 +208,18 @@ public function testEmitPendingPromiseThenFail() {
...
     }
 
     /**
      * @expectedException \Error
      * @expectedExceptionMessage Promise returned from advance() must resolve before calling this method
      */
     public function testGetCurrentBeforeAdvance() {
-        $this->producer->getCurrent();
+        $iterator = $this->producer->iterate();
+        $iterator->getCurrent();
```
kelunik: Why are these test changes necessary?

kelunik: That's a BC break we can't merge then.

trowski: The change is in a trait marked internal. 😛

lib/Emitter.php
```diff
@@ -13,7 +13,7 @@
     private $iterator;
 
     public function __construct() {
-        $this->iterator = new class implements Iterator {
+        $this->iterator = new class {
```
rdlowrey: Just curious on your thought-process behind removing the `implements` expression here (for my own edification)?

trowski: If you look at the new definition of `Internal\Producer`, I would need to add `advance as public; getCurrent as public` for this to still implement `Iterator`.

kelunik:
This breaks BC, we'll revisit it for Amp v3. The corresponding discussion will be kept open and can be found in #168.

PR #199, opened 2018-01-06
Add promise- and coroutine-aware event Mediator

rdlowrey:
This PR proposes a `Promise` and `Coroutine`  aware implementation of the [GoF Mediator pattern](https://en.wikipedia.org/wiki/Mediator_pattern).

### DIFF
--- a/lib/Mediator.php
+++ b/lib/Mediator.php
@@ -35,31 +35,33 @@ class Mediator {
      * @return \Amp\Promise<int> Resolves to the number of subscribers invoked
      */
     public function publish(string $eventName, ...$data): Promise {
-        if (empty($this->eventSubscriberMap[$eventName])) {
-            return new Success(0);
+        return empty($this->eventSubscriberMap[$eventName])
+            ? new Success(0)
+            : new Coroutine($this->doPublish($eventName, $data))
+        ;
+    }
+
+    private function doPublish(string $eventName, array $data): \Generator {
+        foreach ($this->eventSubscriberMap[$eventName] as $id => $callback) {
+            $promises[$id] = call($callback, ...$data);
         }
-        return new Coroutine((function () use ($eventName, $data) {
-            foreach ($this->eventSubscriberMap[$eventName] as $id => $callback) {
-                $promises[$id] = call($callback, ...$data);
+        list($errors, $results) = yield Promise\any($promises);
+        // remove subscribers that returned FALSE (===)
+        foreach ($results as $id => $result) {
+            if ($result === false) {
+                unset($this->eventSubscriberMap[$eventName][$id]);
             }
-            list($errors, $results) = yield Promise\any($promises);
-            // remove subscribers that returned FALSE (===)
-            foreach ($results as $id => $result) {
-                if ($result === false) {
-                    unset($this->eventSubscriberMap[$eventName][$id]);
-                }
-            }
-            // don't leak memory even if it's just an empty array
-            if (empty($this->eventSubscriberMap[$eventName])) {
-                unset($this->eventSubscriberMap[$eventName]);
-            }
-            if ($errors) {
-                throw new MultiReasonException(
-                    $errors,
-                    "Event subscriber(s) threw uncaught exceptions while reacting to {$eventName}"
-                );
-            }
-            return \count($results);
-        })());
+        }
+        // don't leak memory even if it's just an empty array
+        if (empty($this->eventSubscriberMap[$eventName])) {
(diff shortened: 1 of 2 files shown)


kelunik:
I think this should be in its own package.

lib/Mediator.php
```diff
@@ -0,0 +1,83 @@
...
+     * @param string $eventName The name of the event to publish
+     * @param mixed $data Data associated with the event
+     * @return \Amp\Promise<int> Resolves to the number of subscribers invoked
+     */
+    public function publish(string $eventName, ...$data): Promise
+    {
+        if (empty($this->eventSubscriberMap[$eventName])) {
+            return new Success(0);
+        }
+        return new Coroutine((function() use ($eventName, $data) {
```
kelunik: Just use `Amp\call` instead.

lib/Mediator.php
```diff
@@ -0,0 +1,83 @@
...
+            $errors = [];
+            $executionCount = 0;
+            $callbacks = $this->eventSubscriberMap[$eventName];
+            foreach ($callbacks as $id => $callback) {
+                try {
+                    // detach callback if it returns false (===)
+                    if (false === yield call($callback, ...$data)) {
```
kelunik: Any reason why these are executed sequentially?

rdlowrey: General programming rust on my part -- didn't even think about doing it concurrently. Will update.

rdlowrey:
Re: own package ... I thought about it but it was a lot of overhead for 50 LoC without comments. In any case, I just wanted to make a PR for discussion. Happy to create a new repo if folks want that once details are worked out.

lib/Mediator.php
```diff
@@ -0,0 +1,73 @@
...
+                throw new MultiReasonException(
+                    $errors,
+                    "Event subscriber(s) threw uncaught exceptions while reacting to {$eventName}"
+                );
+            }
+            return \count($results);
+        })());
+    }
+
+    public function __debugInfo() {
```
bwoebi: I think this is a bad idea. You usually want to know _what_ subscribed (to track down unexpected invocations of something or see whether something isn't subscribed). Once you have multiple subscribers to an event, a bare number won't tell you anything.

bwoebi:
I agree with @rdlowrey on not using an own package - this just leads to package explosion - and minimal static APIs like this won't ever need to change without fundamental changed to this repo itself. There's like no room for altering this API. It just makes no sense to have this independent from this repo.

kelunik:
That's what we thought for `Amp\Iterator` as well... I think this API will see much lower usage as Amp itself, thus way fewer packages need an update in case of a BC break.

lib/Mediator.php
rdlowrey: Fair. Removing.

test/MediatorTest.php
```diff
@@ -81,34 +81,4 @@ public function testMultiSubscriberErrorCase() {
         $this->assertTrue($caughtExpected);
         $this->assertSame(2, $invocations);
     }
-
-    public function testSubscriberArrayRemovedWhenEmptyToAvoidLeakage() {
```
bwoebi: There's no reason to remove the tests. You are in possession of the class code, so you may as well just ignore the encapsulation here and the read private property here. (`$events = (function() { return $this->eventSubscriberMap; })->call($mediator);`)

Because that's essentially what you want to test: whether the internal state is correct.

rdlowrey: Good call. Updated 👍

joshdifabio:
> ... There's like no room for altering this API. It just makes no sense to have this independent from this repo.

I think that is way overconfident, especially since there is already debate in this thread about changing the implementation.

bwoebi:
Possibly it's a better idea to just commit it to aerys first and then, once the API has matured, copy it here, if we wish to use it globally.

joshdifabio:
> Possibly it's a better idea to just commit it to aerys first and then, once the API has matured, copy it here, if we wish to use it globally.

I think that's a good idea if this would be helpful in Aerys right away, assuming it would live in `Aerys\Internal`. It's hard to have much confidence in an API until it's been used a few times.

brstgt:
Sth different: Isn't sth like 'EventBus' or 'EventDispatcher' a more common name for that than 'Mediator'? See https://github.com/igorw/evenement

kelunik:
I'm closing this PR as there doesn't seem to be any activity. IMO this should be a separate package.

PR #208, opened 2018-03-20
Add Loop::now()

trowski:
We frequently use `microtime()` or `time()` to keep track of last activity time, update time, etc. These system calls are relatively expensive. `Loop::now()` updates only on the first call in a tick (or in `UvLoop` the function `uv_now()` is used).

### DIFF
--- a/lib/Loop/Driver.php
+++ b/lib/Loop/Driver.php
@@ -591,6 +591,20 @@ abstract class Driver
     }
 
     /**
+     * Returns the current loop time in millisecond increments. Note this value does not necessarily correlate to
+     * wall-clock time, rather the value returned is meant to be used in relative comparisons to prior values returned
+     * by this method (intervals, expiration calculations, etc.) and is only updated once per loop tick.
+     *
+     * Extending classes should override this function to return a value cached once per loop tick.
+     *
+     * @return int
+     */
+    public function now(): int
+    {
+        return \microtime(true) * self::MILLISEC_PER_SEC;
+    }
+
+    /**
      * Get the underlying loop handle.
      *
      * Example: the `uv_loop` resource for `libuv` or the `EvLoop` object for `libev` or `null` for a native driver.
--- a/lib/Loop/EvDriver.php
+++ b/lib/Loop/EvDriver.php
@@ -31,7 +31,7 @@ class EvDriver extends Driver
     private $signals = [];
 
     /** @var int Internal timestamp for now. */
-    private $now = 0;
+    private $now;
 
     /** @var bool */
     private $nowUpdateNeeded = false;
@@ -43,6 +43,8 @@ class EvDriver extends Driver
     {
         $this->handle = new \EvLoop;
         $this->nowOffset = (int) (\microtime(true) * self::MILLISEC_PER_SEC);
+        $this->now = \random_int(0, $this->nowOffset);
+        $this->nowOffset -= $this->now;
 
         if (self::$activeSignals === null) {
             self::$activeSignals = &$this->signals;
--- a/lib/Loop/EvDriver.php
+++ b/lib/Loop/EvDriver.php
@@ -11,22 +11,38 @@ class EvDriver extends Driver
 {
     /** @var \EvSignal[]|null */
     private static $activeSignals;
+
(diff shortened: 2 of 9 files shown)


lib/Loop/UvDriver.php
```diff
@@ -160,6 +160,13 @@ public static function isSupported(): bool {
         return \extension_loaded("uv");
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function now(): int {
+        return \uv_now($this->handle);
```
kelunik: `uv_now` is a monotonic timer and not related to the current real time.

kelunik:
The event loop should use a monotonic timer for now instead of `microtime()`. In PHP 7.3 this will be possible without extensions using `hrtime()`. This doesn't avoid any `microtime()` calls then.

lib/Loop/EvDriver.php
```diff
@@ -10,21 +10,34 @@
...
     private $signalCallback;
+
     /** @var \EvSignal[] */
     private $signals = [];
 
+    /** @var int Internal timestamp for now. */
+    private $now;
+
+    /** @var bool */
+    private $nowUpdateNeeded = true;
```
staabm: in the other loop, this defaults to `false`

lib/Loop/UvDriver.php
trowski: Right,that was my intention that will need to be clear through documentation, that `Loop::now()` is not necessarily wall time, but the loop time in milliseconds.

trowski:
@kelunik Yes, I should have stated that `Loop::now()` is not meant to be wall time, but only a relative loop time. As implemented here it's usually wall time, but we can improve it in 7.3 with `hrtime()`.

lib/Loop/EvDriver.php
trowski: Oops, it should be the same to be consistent, though the initial value doesn't matter in this loop.

lib/Loop/Driver.php
```diff
@@ -572,6 +572,13 @@ protected function error(\Throwable $exception) {
         ($this->errorHandler)($exception);
     }
 
+    /**
+     * Get the current loop time in milliseconds.
+     *
+     * @return int
+     */
+    abstract public function now(): int;
```
kelunik: This is a BC break.

trowski: Ugh… I suppose so. We could add an `ExtendedDriver`, but I'd rather wait for v3 then.

staabm: What about adding the method in the implementation classes but not in the interface?

kelunik: @staabm Then it can't be used, because the interface doesn't cover it.

staabm: Users which dont need to abstract the impl and use it at app level could use it.

Inside amp you couldn‘t

kelunik: The function is mostly useful in libraries, not in applications, so...

bwoebi:
Not sure why this needs to be abstract. Can't we just simply use the `NativeLoop::now()` implementation within `Driver` and just simply override it within the individual drivers?

`NativeLoop::now()` should be a good enough fallback for any case, we just have additional guarantees about monotonicity within other Drivers.

lib/Loop/EvDriver.php
```diff
@@ -11,22 +11,38 @@ class EvDriver extends Driver
...
+    /** @var bool */
+    private $nowUpdateNeeded = false;
+
+    /** @var int Loop time offset from microtime() */
+    private $nowOffset;
+
     public function __construct()
     {
         $this->handle = new \EvLoop;
+        $this->nowOffset = (int) (\microtime(true) * self::MILLISEC_PER_SEC);
```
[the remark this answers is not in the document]
trowski: Done.

lib/Loop/EvDriver.php
```diff
@@ -11,22 +11,40 @@ class EvDriver extends Driver
...
+        $this->now = \random_int(0, $this->nowOffset);
```
kelunik: Any reason why `$this->now` is updated here?

PR #215, opened 2018-04-08
Alternative approach to adding Loop::now()

trowski:
This PR (hopefully) adds `Loop::now()` in a BC way.


bwoebi:
Looks good, but aren't we going to use uv_now() with UvDriver?

trowski:
@bwoebi Yep, I forgot about `uv_now()`. Pushed a new commit that uses it.

kelunik:
I just noticed that this breaks decorating driver implementations, as they have the inherited now() implementation then, but never call tick().

kelunik:
Closing in favor of #208.

PR #216, opened 2018-04-08
Fix 212 and ignore unreference on invalid watchers

trowski:
Putting this in the same PR as they're *somewhat* related. I can drop the second commit if we decide we don't want it.

Context for ignoring unreference on invalid loop watchers: [`Processor::ready()` in amphp/mysql](https://github.com/amphp/mysql/blob/f3e1f159c971cd23f87014cb97601d3deae321ec/src/Internal/Processor.php#L172-L177)

When a result or statement is destroyed, the connection is marked as ready. During shutdown, this results in `Socket::unreference()` potentially being called after `Socket::__destruct()` has been called.

### DIFF
--- a/lib/Loop.php
+++ b/lib/Loop.php
@@ -247,6 +247,12 @@ final class Loop {
      * @return void
      */
     public static function disable(string $watcherId) {
+        if (!isset(self::$driver)) {
+            // Prior to PHP 7.2, self::$driver may be unset during destruct.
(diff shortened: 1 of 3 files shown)


bwoebi:
I recommend adding `PHP_VERSION_ID < 70200 && ` in the if conditions, to make sure that it won't slip over our heads if there are future problems with this.

trowski:
@bwoebi Added the version check to the conditionals.

PR #219, opened 2018-04-21
Fix EventDriver garbage collection

### LINKED ISSUE #177: Not possible to cleanly shutdown EventDriver
When a loop is set, the old loop cannot be cleanly shutdown.
Tested with EventDriver, the driver is not destructed if there are registered watchers. The API does not allow a manual shutdown. This also leads to a memleak as the object is never GCed.

See here for a test and a change that fixes the issue (quick + dirty):
https://gist.github.com/brstgt/0a83963136436ef5e266c1d807e513c4

kelunik:
Cyclic references prevented the garbage collection. I guess this is a bug in the extension, but we can work around it by avoiding these references.

Fixes #177.

### DIFF
--- a/lib/Loop/EventDriver.php
+++ b/lib/Loop/EventDriver.php
@@ -29,11 +29,20 @@ class EventDriver extends Driver {
         $this->handle = new \EventBase;
         $this->now = (int) (\microtime(true) * self::MILLISEC_PER_SEC);
 
+        $errorHandler = &$this->errorHandler;
+        $error = static function (\Throwable $exception) use (&$errorHandler) {
+            if ($errorHandler === null) {
+                throw $exception;
+            }
+
+            $errorHandler($exception);
+        };
+
         if (self::$activeSignals === null) {
             self::$activeSignals = &$this->signals;
         }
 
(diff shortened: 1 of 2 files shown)


lib/Loop/EventDriver.php
```diff
@@ -49,15 +58,19 @@ public function __construct() {
...
+                $error($exception);
             }
         };
 
-        $this->timerCallback = function ($resource, $what, Watcher $watcher) {
+        $events = $this->events;
+        $this->timerCallback = static function ($resource, $what, Watcher $watcher) use ($error, $events) {
             if ($watcher->type & Watcher::DELAY) {
-                $this->cancel($watcher->id);
+                if (isset($events[$watcher->id])) {
```
trowski: `Driver::cancel()` needs to be called here.

bwoebi:
We should do this anyway for performance reasons, otherwise Closures like these will clutter the gc root buffer.

trowski:
@bwoebi Only three closures are created for every instance. I'm still confused as to why loops are being swapped out so often that this is even a problem.

kelunik:
Closing, because it will be fixed in the extension.

PR #226, opened 2018-05-12
Document yield VS yield from

prolic:


docs/coroutines/README.md
```diff
@@ -44,3 +44,58 @@ All `yield`s in a coroutine must be one of the following three types:
 | `Amp\Promise` | Any promise instance may be yielded and control will be returned to the coroutine once the promise resolves. If resolution fails the relevant exception is thrown into the generator and must be handled by the application or it will bubble up. If resolution succeeds the promise's resolved value is sent back into the generator. |
 | `React\Promise\PromiseInterface` | Same as `Amp\Promise`. Any React promise will automatically be adapted to an Amp promise. |
 | `array` | Yielding an array of promises combines them implicitly using `Amp\Promise\all()`. An array with elements not being promises will result in an `Amp\InvalidYieldError`. |
+
+## Yield VS Yield From
+
+Yield promises from within a `\Generator`, the code within the `\Generator` will continue, as soon as the promise is resolved. Use `yield from` to yield the result of a `\Generator`. Instead of using yield from, you can also use `yield new Coroutine($this->bar());` or `yield call([$this, "bar"]);`.
```
kelunik: I think it should mention _delegation_ in combination with _yield from_.

docs/coroutines/README.md
```diff
@@ -44,3 +44,58 @@ All `yield`s in a coroutine must be one of the following three types:
...
+
+An example:
+
+```php
+class Foo
+{
+    public function foo1(): \Amp\Promise
```
kelunik: I'd remove the leading backslashes for improved readability.

docs/coroutines/README.md
```diff
@@ -44,3 +44,58 @@ All `yield`s in a coroutine must be one of the following three types:
...
+    {
+        return new \Amp\Coroutine($this->bar());
+    }
+
+    public function foo2(): \Amp\Promise
```
kelunik: I guess these should have better names or a comment that describes what they do. Maybe something like _delegationWithYieldFrom_?

docs/coroutines/README.md
```diff
@@ -44,3 +44,58 @@ All `yield`s in a coroutine must be one of the following three types:
...
+
+Outputs:
+```
+int(3)
+int(3)
+int(3)
+```
+
+{:.note}
+Yield is used to "await" promises, yield from can be used to delegate to a sub-routine. Yield from should only be used to delegate to private methods, any public API should always return promises instead of generators.
```
kelunik: I think this text should be right at the top of the section.

prolic:
@kelunik updated

prolic:
BTW: yield from -> what is this from? a PHP keyword? Can't find it in the manual

jvasseur:
@prolic `yield from` is documented here : http://php.net/manual/en/language.generators.syntax.php#control-structures.yield.from

prolic:
thanks @kelunik @jvasseur I added that link to the docs as well

kelunik:
Thanks!

PR #234, opened 2018-07-27
add Promise\wrap

prolic:


bwoebi:
Hm, we had quite a similar function in Amp v1: `\Amp\pipe`: https://github.com/amphp/amp/blob/1.x/lib/functions.php#L597-L639

We removed it in v2 in favor of generators; i.e.:

```php
\Amp\call(function() use ($promise) {
    return doSomeTransformation(yield $promise);
});
```

So I guess we won't accept this either, for the same reason.

prolic:
@bwoebi thanks, i like your suggestion even more

kelunik:
@bwoebi This is different from pipe, as it also gets executed on errors. I'd probably merge it.

bwoebi:
@kelunik Sure, but you can easily try/catch that in a coroutine, if you wish error handling - hence "similar".

lib/functions.php
```diff
@@ -441,6 +441,34 @@ function some(array $promises, int $required = 1): Promise
...
     }
+
+    /**
+     * Wraps a promise into another promise, altering the exception or result
+     *
+     * @param Promise $promise
+     * @param callable $callback
+     * @return Promise
+     */
+    function wrap(Promise $promise, callable $callback): Promise
```
kelunik: Could you please adjust this PR to accept a ReactPHP promise like all other functions?

prolic: done

kelunik: A passed ReactPHP promise needs to be adapted before `onResolve()` can be called.

prolic: ah totally missed that, my bad :)

PR #239, opened 2018-10-05
Add Iterator\collect

kelunik:
Another possible name would be `toArray`, not sure which one is better.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -451,6 +451,7 @@ namespace Amp\Iterator
     use Amp\Iterator;
     use Amp\Producer;
     use Amp\Promise;
+    use function Amp\call;
     use function Amp\coroutine;
     use function Amp\Internal\createTypeError;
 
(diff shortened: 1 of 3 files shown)


bwoebi:
I think `toArray` is favorable here as it's more explicit. And is also alluding to the well known `iterator_to_array()`.

trowski:
I also prefer `toArray`. `collect` is elegant, but not self-documenting like `toArray`.

Sevavietl:
@trowski Hi.

Will this PR be merged soon? Need this function in multiple projects, so do not want to copy-paste my own implementation.

Thank you in advance.

kelunik:
@Sevavietl now. ;-)

Sevavietl:
@kelunik Thank you for the quick response.

Do you know when the next release will be? So I can require it instead of dev-master.

kelunik:
Soon.™

I planned today, not sure whether I'll get to that.

Sevavietl:
@kelunik Thank you. No biggie.

By the way, I have created [some custom matchers](https://github.com/OWOX/PhpSpecAmp) for phpspec, so as soon as this function will be released will use it for spec-testing of `Producers` and `Iterators`.

Thank you again for the great library.

kelunik:
@Sevavietl Thanks!

PR #240, opened 2018-10-05
Simplify Iterator\concat implementation

kelunik:
I'm not sure whether the new implementation matches the old one 100% in behavior, but I think it does.

The fact that all remaining iterators are iterated even if one of the previous iterators failed seems questionable to me, but I guess it can't be fixed until a new major is released.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -571,49 +571,27 @@ namespace Amp\Iterator
             }
         }
 
-        $emitter = new Emitter;
-        $previous = [];
-        $promise = Promise\all($previous);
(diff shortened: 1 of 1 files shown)


trowski:
It's slightly different in that the original implementation beings iterating all iterators simultaneously. Not sure how much difference that makes, but that sort of explains the other odd behavior of continuing to iterate after one fails.

If that change doesn't matter, this looks good.

kelunik:
Let's not introduce minor changes here. If we fix it, it won't iterate the remaining iterators in case of an exception.

Issue #244, opened 2018-10-27
CountdownEvent & ManualResetEventSlim


prolic:
I got this idea from C#, where it is used for threads, however I found it quite useful for amp as well.

See:
https://github.com/prooph/event-store-client/blob/master/tests/CountdownEvent.php
and
https://github.com/prooph/event-store-client/blob/master/src/Internal/ManualResetEventSlim.php

if you like this idea, I can provide a PR.

prolic:
@bwoebi about ManualResetEventSlim: Basically yes, but it removes a lot of boilerplate to write.

F.e. in a subscription you can do this in constructor:

`$this->stopped = new ManualResetEventSlim(true);`

then when subscription was dropped this:

`$this->stopped->set();`

when start() is called:

`$this->stopped->reset();`

and when you call stop(int $timeout) you do this:

`return $this->stopped->wait($timeout);`

The set / reset methods are easily replaced with resetting a deferred and resolving with true value, so there is not `that` much benefit, but the wait method is really some boilerplate to write everywhere.

To me it is also very useful in testing, because I don't need to write this `wait` boilerplate all over the tests again and again.

However, if you don't want to have it, I'm okay with providing a PR only for CountdownEvent - do you like the implementation as is or do you want any changes before before I send the PR?

bwoebi:
I do agree that wait() is quite a bit of boilerplate, but the main issue is that we have no timeout function which simply resolves with a default value and instead always throws.

I think it could be worth adding a timeout function to amp, which simply resolves to a deferred with a value (given as argument) instead of throwing. I've also missed that multiple times and ended up writing a boilerplate one-off Loop::delay() instead.

This would also effectively reduce your wait boilerplate to `timeoutWithDefault($this->stopped->promise(), false)`.

prolic:
Okay, sounds good, I will provide a PR this weekend.
Sascha-Oliver Prolic


Am Mi., 31. Okt. 2018 um 17:47 Uhr schrieb Bob Weinand <
notifications@github.com>:

kelunik:
Closing, as https://github.com/amphp/sync/pull/9 and https://github.com/amphp/amp/pull/245 have been merged.

PR #245, opened 2018-11-02
add timeoutWithDefault function

prolic:
see https://github.com/amphp/amp/issues/244

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -248,20 +248,23 @@ namespace Amp\Promise
             }
         }
 
-        $promise = timeout($promise, $timeout);
+        $deferred = new Deferred;
 
-        $deferred = new Deferred();
-        $newPromise = $deferred->promise();
+        $watcher = Loop::delay($timeout, function () use (&$deferred, $default) {
+            $temp = $deferred; // prevent double resolve
+            $deferred = null;
+            $temp->resolve($default);
+        });
+        Loop::unreference($watcher);
 
-        $promise->onResolve(function ($exception, $value) use ($deferred, $default) {
-            if ($exception) {
-                $deferred->resolve($default);
-            } else {
-                $deferred->resolve($value);
+        $promise->onResolve(function ($exception) use (&$deferred, $promise, $watcher) {
+            if ($deferred !== null) {
+                Loop::cancel($watcher);
+                $deferred->resolve($promise);
             }
         });
 
-        return $newPromise;
+        return $deferred->promise();
     }
 
     /**
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -95,7 +95,6 @@ namespace Amp
 namespace Amp\Promise
 {
 
-    use function Amp\call;
     use Amp\Deferred;
     use Amp\Failure;
     use Amp\Loop;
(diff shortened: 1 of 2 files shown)


lib/functions.php
```diff
@@ -223,6 +223,47 @@ function timeout($promise, int $timeout): Promise
...
+        }
+
+        $promise = timeout($promise, $timeout);
+
+        $deferred = new Deferred();
+        $newPromise = $deferred->promise();
+
+        $promise->onResolve(function ($exception, $value) use ($deferred, $default) {
+            if ($exception) {
+                $deferred->resolve($default);
```
kelunik: An exception should probably be thrown here instead of resulting in the default value.

prolic: You're right, give me a minute.

kelunik: I wouldn't make use of the timeout function, but rather adapt the code of it to suit this function.

prolic: No wait, that's the whole purpose of the default value. You timeout with a default and when you'll receive a timeout exception, you'll get the default value.

kelunik: You'll get the default on a timeout, but not if the promise fails?

prolic:
@kelunik updated

trowski:
This is way less code:

```php
function timeoutWithDefault($promise, int $timeout, $default = null): Promise
{
    return call(function () use ($promise, $timeout, $default) {
        try {
            return yield timeout($promise, $timeout);
       } catch (TimeoutException $excetpion) {
            return $default;
       }
    });
}
```

This is so little code that I question the need for this, but also so little that I don't really care maintance-wise either.

prolic:
@trowski: @kelunik asked me not to reuse the timeout function. I know it's not much code, but having those 7 lines of code 50 times in your code base really sucks :)

trowski:
I think mostly because you originally didn't use `timeout()` within a coroutine, which then makes it nearly as complicated. Otherwise I see no reason to not reuse the function.

You can define functions in your code base you know. :wink: But yes, I can see where this can be useful for many people, so let's add it.

test/TimeoutWithDefaultTest.php
```diff
@@ -0,0 +1,132 @@
...
+                $result = $value;
+            };
+
+            $promise->onResolve($callback);
+
+            $this->assertSame($value, $result);
+        });
+    }
+
+    public function testNonPromise()
```
prolic: @trowski I tested your suggested implementation and then this test fails, because the TypeError is not thrown.

trowski: It should be…

trowski: Oh, right, because it's not resolving the promise with the TypeError. Fine, then it needs to be this way:

```php
function timeoutWithDefault($promise, int $timeout, $default = null): Promise
{
    $promise = timeout($promise, $timeout);

    return call(function () use ($promise, $default) {
        try {
            return yield $promise;
        } catch (TimeoutException $excetpion) {
            return $default;
        }
    });
}

```

lib/functions.php
```diff
@@ -240,31 +244,15 @@ function timeout($promise, int $timeout): Promise
...
-        Loop::unreference($watcher);
-
-        $promise->onResolve(function ($exception) use (&$deferred, $promise, $watcher) {
-            if ($deferred !== null) {
-                Loop::cancel($watcher);
-                $deferred->resolve($promise);
+        return call(function () use ($promise, $default) {
+            try {
+                return yield $promise;
+            } catch (TimeoutException $excetpion) {
```
trowski: Totally my fault, but typo: $excetpion :smile:

trowski:
The style fixer is going to yell at you. 🙄
Otherwise looks good.

prolic:
done

PR #246, opened 2018-11-02
add CountdownBarrier

prolic:
see https://github.com/amphp/amp/issues/244

### DIFF
--- a/lib/CountdownBarrier.php
+++ b/lib/CountdownBarrier.php
@@ -6,17 +6,17 @@ use InvalidArgumentException;
 use RuntimeException;
 
 /**
- * A countdown event is a promise that is resolved when it was signaled n times
+ * A countdown event is a promise that is resolved when it was signaled n times.
  *
  * **Example**
  *
  * ```php
- * $countdownEvent = new CountdownEvent(2);
+ * $countdownEvent = new \Amp\CountdownBarrier(2);
  * $countdownEvent->signal();
  * $countdownEvent->signal(); // promise is now resolved
  * ```
  */
-final class CountdownEvent implements Promise
+final class CountdownBarrier
 {
     /** @var int */
     private $counter;
@@ -50,10 +50,10 @@ final class CountdownEvent implements Promise
     }
 
     /**
-     * {@inheritdoc}
+     * @return \Amp\Promise
      */
-    public function onResolve(callable $onResolved)
+    public function promise(): Promise
     {
-        $this->deferred->promise()->onResolve($onResolved);
+        return $this->deferred->promise();
     }
 }
--- a/lib/CountdownBarrier.php
+++ b/lib/CountdownBarrier.php
@@ -11,41 +11,96 @@ use RuntimeException;
  * **Example**
  *
  * ```php
- * $countdownEvent = new \Amp\CountdownBarrier(2);
- * $countdownEvent->signal();
- * $countdownEvent->signal(); // promise is now resolved
+ * $countdownBarrier = new \Amp\CountdownBarrier(2);
+ * $countdownBarrier->signal();
+ * $countdownBarrier->signal(); // promise is now resolved
  * ```
  */
 final class CountdownBarrier
 {
     /** @var int */
-    private $counter;
+    private $initialCount;
+    /** @var int */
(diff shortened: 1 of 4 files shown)


lib/CountdownEvent.php
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * **Example**
+ *
+ * ```php
+ * $countdownEvent = new CountdownEvent(2);
+ * $countdownEvent->signal();
+ * $countdownEvent->signal(); // promise is now resolved
+ * ```
+ */
+final class CountdownEvent implements Promise
```
kelunik: I don't really like the name event for something that implements Promise. Maybe CountdownBarrier?

lib/CountdownEvent.php
```diff
@@ -0,0 +1,59 @@
...
+        if (0 === $this->counter) {
+            throw new RuntimeException('CountdownEvent already resolved');
+        }
+
+        --$this->counter;
+
+        if (0 === $this->counter) {
+            $this->deferred->resolve(true);
+        }
+    }
```
kelunik: We usually separate the objects that resolve a promise and the object that is passed to the API user to subscribe to.

prolic: So what would you suggest here instead?

kelunik: Having a `promise()` method like `Deferred` that returns the promise to the consumer. Or maybe it's fine here, since these objects aren't supposed to be passed across boundaries? Dunno.

lib/CountdownEvent.php
kelunik: Barrier isn't the right name either, because that'd mean all `signal()` calls return promises that resolve once the count is 0.

prolic:
@kelunik done

lib/CountdownEvent.php
prolic: A great, I just renamed it to Barrier, haha.
For people coming from .NET CountdownEvent is definitely a suitable name, I don't know what describes it better.

trowski:
CountdownBarrier makes sense, it doesn't imply that `signal()` should return a promise IMO.

Thoughts/suggestions:
- Return a `bool` from `signal()` indicating if the count reached 0, resolving the promise.
- Add a method `add(int $count): void`? Throw if resolved or `$count <= 0`.
- Add a method `reset(int $count = null): void`. Use the initial count if `$count === null`.
- Add a `$count` argument to `signal()` with a default of 1. Throw if <= 0.

prolic:
@trowski done

prolic:
@trowski done

lib/CountdownBarrier.php
```diff
@@ -0,0 +1,111 @@
...
+            throw new Error('CountdownBarrier already resolved');
+        }
+
+        $this->currentCount -= $signalCount;
+
+        if (0 === $this->currentCount) {
+            $this->deferred->resolve(true);
+
+            return true;
+        }
```
kelunik: This should check for `$this->currentCount < 0`  and throw an error then, too.

trowski: Throw or just make it 0?

trowski:
How about `increase()`/`decrease()` instead of `addCount()`/`signal()`?

kelunik:
@prolic Do you have time to join `#amphp` on Freenode right now?

kelunik:
@prolic Do all signalers usually wait for all others or is it just one different execution flow that waits?

prolic:
Usually you have one execution flow that waits and one or more that are
sending signals.

prolic:
Too bad this was not in 2.1 release. Any problems with this?

kelunik:
We want to give it some more thought. We can always release v2.2.0.

prolic:
@kelunik It's more than a year later now. Are you still interested in this? Otherwise I might release it as an independent open source package.

kelunik:
@prolic I guess `amphp/sync` might be a better fit for this than `amphp/amp` itself. What do you think?

prolic:
Sure why not. Should I close this PR and submit the files to that other repo?

kelunik:
@prolic If you have time for that, that'd be great!

PR #250, opened 2018-11-29
Add Flow

trowski:
`Flow` is a replacement for `Iterator`, providing a simpler interface, the ability to emit `<value,key>` pairs, and notification to the emitter if all consumers have been destroyed.

```php
use Amp\AsyncGenerator;
use Amp\Delayed;
use Amp\Loop;

Loop::run(function () {
    try {
        $generator = new AsyncGenerator(function (callable $yield) {
            yield $yield(1);
            yield $yield(yield new Delayed(500, 2));
            yield $yield(3);
            yield $yield(yield new Delayed(300, 4));
            yield $yield(5);
            yield $yield(6);
            yield $yield(yield new Delayed(1000, 7));
            yield $yield(8);
            yield $yield(9);
            yield $yield(yield new Delayed(600, 10));
        });

        while ([$value, $key] = yield $generator->continue()) {
            \printf("AsyncGenerator yielded <Key: %d, Value: %d>\n", $key, $value);
            yield new Delayed(100); // Listener consumption takes 100 ms.
        }
    } catch (\Exception $exception) {
        \printf("Exception: %s\n", $exception);
    }
});
```

I'm opening this for discussion. This PR is not ready to be merged as-is (more tests are needed, better documentation, etc.)

### DIFF
--- a/lib/Flow.php
+++ b/lib/Flow.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Amp;
+
+/**
+ * Defines a flow, an asynchronous generator that yields key/value pairs as data is available.
+ */
+interface Flow
+{
+    /**
+     * Succeeds with a [value, key] pair or null if no more values are available. If the flow fails, the returned promise
+     * will fail with the same exception.
+     *
+     * @return \Amp\Promise<[mixed $value, mixed $key]|null>
+     *
+     * @throws \Error If the prior promise returned from this method has not resolved.
+     * @throws \Throwable The exception used to fail the flow.
+     */
+    public function continue(): Promise;
+}
--- a/lib/AsyncGenerator.php
+++ b/lib/AsyncGenerator.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Amp;
+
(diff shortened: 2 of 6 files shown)


trowski:
@kelunik Part of the reason for supporting keys was then when `null` is resolved from `Flow::continue()`, you know you've reached the end since a `<value,key>` pair was not emitted. Another option would be to just not allow emitting `null`, which would be reasonable IMO. Thoughts?

lib/Flow.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+interface Flow
+{
+    /**
+     * Succeeds with a [value, key] pair or null if no more values are available. If the flow fails, the returned promise
+     * will fail with the same exception.
+     *
+     * @return \Amp\Promise<[mixed $value, mixed $key]|null>
+     *
+     * @throws \Error If the prior promise returned from this method has not resolved.
```
kelunik, line 16: I think we should allow concurrent `continue` operations in `Flow`, as that's another issue we currently have with `Iterator`.

trowski:
Closing in favor of #316.

PR #260, opened 2019-02-05
fix phpdoc types

mmasiukevich:
Incorrect types were indicated, due to which static analysis led to false positives.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -49,7 +49,7 @@ namespace Amp
      * Calls the given function, always returning a promise. If the function returns a Generator, it will be run as a
      * coroutine. If the function throws, a failed promise will be returned.
      *
-     * @param callable(mixed ...$args): mixed $callback
+     * @param callable(mixed ...$args):\Generator|mixed $callback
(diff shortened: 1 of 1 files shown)


mmasiukevich:
Not now ... there are more problems with the type description

kelunik:
Seems like `mixed` should indeed be used here, I'll re-open the PR if you're fine with that.

mmasiukevich:
Yes of course. There just is 1 more commit: https://github.com/mmasiukevich/amp/commit/d56be41c5aad8a9e402752a94534fb113131e521
So the analyzer (psalm) is satisfied.

mmasiukevich:
From the point of view of analyzers, it would be nice to support the @template annotation.
psalm issue: https://github.com/vimeo/psalm/issues/1283  (@template support by phan & psalm)

muglug:
@mmasiukevich if you add, make sure to use `@psalm-return` and `@psalm-param` so as not to confuse other static analysers that don't understand `@template`

kelunik:
Please open an issue if you want to discuss psalm integration.

PR #271, opened 2019-05-13
Add failing Coroutine test

kelunik:
I have currently no idea how that exception can even be caught in the test case.

### DIFF
--- a/lib/Coroutine.php
+++ b/lib/Coroutine.php
@@ -144,15 +144,4 @@ final class Coroutine implements Promise
             });
         }
     }
-
-    public function __destruct()
-    {
-        try {
-            $this->result = null;
-        } catch (\Throwable $e) {
-            Loop::defer(static function () use ($e) {
-                throw $e;
-            });
-        }
-    }
 }
--- a/lib/Coroutine.php
+++ b/lib/Coroutine.php
@@ -72,20 +72,20 @@ final class Coroutine implements Promise
             return;
         }
 
-        /** @var bool Used to control iterative coroutine continuation. */
-        $immediate = true;
-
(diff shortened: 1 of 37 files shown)


lib/Coroutine.php
```diff
@@ -15,126 +15,150 @@ final class Coroutine implements Promise
...
+                    $immediate = false;
+                    $yielded->onResolve($onResolve);
+                } while ($immediate);
 
-                $this->immediate = true;
+                $immediate = true;
             } catch (\Throwable $exception) {
-                $this->fail($exception);
-                $this->onResolve = null;
+                try {
```
trowski: Can we wrap the `try` started on line 97 in a single try/catch rather than separate try/catch blocks in catch and finally?

lib/Coroutine.php
```diff
@@ -15,126 +15,150 @@ final class Coroutine implements Promise
...
-    /**
-     * Attempts to transform the non-promise yielded from the generator into a promise, otherwise returns an instance
-     * `Amp\Failure` failed with an instance of `Amp\InvalidYieldError`.
-     *
-     * @param mixed $yielded Non-promise yielded from generator.
-     *
-     * @return \Amp\Promise
-     */
-    private function transform($yielded): Promise
+    public function __destruct()
```
trowski: Should this destructor be part of `Placeholder`?

lib/Coroutine.php
kelunik: Done.

lib/Coroutine.php
kelunik: Done.

PR #273, opened 2019-05-31
Support 32-bit platforms

kelunik:
It will still wrap around every ~ 24 days, but not randomly. At least not until 2038.


trowski:
Your way is better, manually merged.

Issue #264, opened 2019-02-25
Support for pcntl_async_signals in NativeDriver


Ekstazi:
### Php 7.1 introduce new way to handle signals.

If pcntl_async_signals is set to true then all signals will be processed async and immediate.

But if context of execution is not in NativeDriver (in synchronous code) then this error happens:
```
^CPHP Warning:  Invalid callback Amp\Loop\NativeDriver::handleSignal, cannot access private method Amp\Loop\NativeDriver::handleSignal() in /app/test.php on line 9
PHP Stack trace:
PHP   1. {main}() /app/test.php:0
PHP   2. Amp\Loop::run($callback = *uninitialized*) /app/test.php:11
PHP   3. Amp\Loop\NativeDriver->run() /app/vendor/amphp/amp/lib/Loop.php:84
PHP   4. Amp\Loop\NativeDriver->tick() /app/vendor/amphp/amp/lib/Loop/Driver.php:72
PHP   5. Amp\Loop\NativeDriver->dispatch($blocking = *uninitialized*) /app/vendor/amphp/amp/lib/Loop/Driver.php:134
PHP   6. {closure:/app/test.php:8-10}(*uninitialized*, *uninitialized*) /app/vendor/amphp/amp/lib/Loop/NativeDriver.php:130
```

Example code to reproduce error:
```php
<?php
require __DIR__.'/vendor/autoload.php';

pcntl_async_signals(true);
\Amp\Loop::onSignal(SIGINT, function (){
	echo "sigterm\n";
});
\Amp\Loop::repeat(100, function(){
	sleep(1);
});
\Amp\Loop::run();
```

I understand that amp not support sync code and this error will be only in rare cases.

P.S.: sorry for bad english.

Ekstazi:
May be need to enable async signals by default ?

PR #275, opened 2019-06-07
Automatically enable async signal handling when possible

danog:


lib/Loop/NativeDriver.php
```diff
@@ -43,10 +43,14 @@ class NativeDriver extends Driver
...
     private $signalHandling;
 
+    /** @var bool */
+    private $manualSignalHandling;
+
     public function __construct()
     {
         $this->timerQueue = new Internal\TimerQueue;
         $this->signalHandling = \extension_loaded("pcntl");
+        $this->manualSignalHandling = !(\PHP_MAJOR_VERSION >= 7 && \PHP_MINOR_VERSION >= 1);
```
kelunik: Please use `PHP_VERSION_ID < 70100` instead.

danog: Done

danog:
> I don't like such a behavior change. Usually you don't need async signal handling in async applications, because they're already fairly non-blocking and `pcntl_signal_dispatch` gets called often enough.

I just wanted to implement this to reduce the (small, but still noticeable) overhead of calling pcntl_signal_dispatch at each tick; and it doesn't really hurt to reduce overhead by delegating some stuff to PHP, when possible and not too complex.

trowski:
I believe we discussed this in the past and decided it was better to not enable async signal dispatching by default because it removed the guarantee that only a single watcher callback could be invoked at any given time. With async signal dispatch, a signal watcher may be invoked at *any* time, making it even more difficult to reason about execution flow. Since #264 was fixed, you can invoke `pcntl_async_signals` in your application if you want more immediate signal handling, of course with the caveat I mentioned before. We could check if async signal dispatching is enabled before calling `pcntl_signal_dispatch`, but then we're adding another function call to the normal case. Feel free to test, but I'd be surprised if there was a performance benefit.

danog:
Instead of adding an additional function call, we could check if async signal handling is enabled when a new signal handler is set, or maybe add a new method to manually enable/disable signal handling (I prefer the first option).

danog:
Done

trowski:
IMO this is unnecessary. You're adding complexity to avoid a single function call per tick. `pcntl_dispatch_signals` checks a counter for pending signals and immediately returns if there are no pending signals. There is essentially no benefit to be had here.

`pcntl_async_signals` can be toggled at any time, so checking it only on construction or when adding a signal watcher is insufficient.

danog:
While it is true that `pcntl_async_signals` can toggled at any time, it is usually toggled while setting up the signal handlers.

I don't see the point of even using `pcntl_async_signals` (except maybe for a more immediate call of the callback) if the AMPHP event loop already handles signals on its own: with the current unmodified event loop, `pcntl_async_signals` just adds more overall overhead, forcing the use of both PHP's internal signal scheduler and AMPHP's scheduler.

kelunik:
Closing in favor of a more predictable execution order.

PR #278, opened 2019-07-15
Add various sleep functions

danog:


kelunik:
I don't see any value in usleep. I'd rather just have a single delay function accepting milliseconds.

danog:
Yeah I didn't think it'd be particularly useful, but I'd really like a sleep function that behaves like PHP sleep (accepting seconds, even if in floating point)

trowski:
Since Amp uses milliseconds for timers, I also prefer a single sleep function accepting milliseconds. Adding ` * 1000` isn't a big deal if you want seconds. Let's keep the time currency consistent.

danog:
OK then

lib/functions.php
```diff
@@ -90,6 +90,18 @@ function asyncCall(callable $callback, ...$args)
...
     }
+
+    /**
+     * Sleeps for the specified number of milliseconds.
+     *
+     * @param int $milliseconds
+     *
+     * @return Promise
+     */
+    function sleep(int $milliseconds): Promise
```
kelunik: Could you please change the name to `delay` to match `Delayed`?

danog: Done

kelunik: I meant changing the function name, not the return type.

kelunik:
Thanks!

PR #285, opened 2019-10-10
Add Timing API

kelunik:
Relates to https://github.com/amphp/http-client/issues/193.


kelunik:
I won't use this anymore for http-client, given https://github.com/amphp/http-client/pull/220, but we might still want to merge this (in the future).

kelunik:
This isn't something that should be in the base package.

PR #299, opened 2020-02-06
Add parameter for custom timeout message

trowski:
This would be helpful for adding a custom message, like "Connecting to PostgreSQL server timed out."

### DIFF
--- a/lib/TimeoutCancellationToken.php
+++ b/lib/TimeoutCancellationToken.php
@@ -15,14 +15,15 @@ final class TimeoutCancellationToken implements CancellationToken
 
     /**
      * @param int $timeout Milliseconds until cancellation is requested.
+     * @param string $message Message for TimeoutException. Default is "Operation timed out".
      */
-    public function __construct(int $timeout)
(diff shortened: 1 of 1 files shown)


lib/TimeoutCancellationToken.php
```diff
@@ -15,14 +15,15 @@ final class TimeoutCancellationToken implements CancellationToken
...
-    public function __construct(int $timeout)
+    public function __construct(int $timeout, string $message = "Operation timed out")
     {
         $source = new CancellationTokenSource;
         $this->token = $source->getToken();
 
-        $this->watcher = Loop::delay($timeout, static function () use ($source) {
-            $source->cancel(new TimeoutException);
+        $this->watcher = Loop::delay($timeout, static function () use ($source, $message) {
+            $source->cancel(new TimeoutException($message));
```
kelunik: Should it be possible to provide the whole exception instead?

trowski: I don't think I care for that, since then the instance has to be made prior to the timeout (unless we used a callable, but then again, more complexity for something simple).

PR #301, opened 2020-02-25
Improve compatibility with static analysis tools

enumag:
See https://medium.com/@ondrejmirtes/generics-in-php-using-phpdocs-14e7301953

### DIFF
--- a/lib/Promise.php
+++ b/lib/Promise.php
@@ -23,8 +23,8 @@ interface Promise
      * Note: You shouldn't implement this interface yourself. Instead, provide a method that returns a promise for the
      * operation you're implementing. Objects other than pure placeholders implementing it are a very bad idea.
      *
-     * @param callable(\Throwable|null $reason, TValue|null $value) $onResolved `$reason` shall be `null` on
-     *     success, `$value` shall be `null` on failure.
+     * @param callable(\Throwable|null, TValue|null) $onResolved The first argument shall
+     *     be `null` on success, while the second shall be `null` on failure.
      *
      * @return void
      */
--- a/lib/Promise.php
+++ b/lib/Promise.php
@@ -4,6 +4,8 @@ namespace Amp;
 
 /**
  * Representation of the future value of an asynchronous operation.
+ *
+ * @template-covariant TValue
  */
 interface Promise
 {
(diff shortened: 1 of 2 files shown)


lib/Promise.php
```diff
@@ -21,7 +23,7 @@ interface Promise
      * Note: You shouldn't implement this interface yourself. Instead, provide a method that returns a promise for the
      * operation you're implementing. Objects other than pure placeholders implementing it are a very bad idea.
      *
-     * @param callable(\Throwable|null $reason, $value) $onResolved `$reason` shall be `null` on
+     * @param callable(\Throwable|null $reason, TValue|null $value) $onResolved `$reason` shall be `null` on
```
kelunik: Accoding to https://psalm.dev/, there needs to be a space behind `callable`.

enumag: I don't think so. https://psalm.dev/docs/annotating_code/type_syntax/callable_types/

kelunik: In that case it's probably due to `$reason` and `$value` which need to be removed?

enumag: Hmm... not sure how to rewrite the message then: "`$reason` shall be `null` on success, `$value` shall be `null` on failure."

enumag: But I think that PHPStan at least accepts the syntax with names. I don't know about Psalm though since I'm not using it.

lib/Promise.php
```diff
@@ -21,7 +23,7 @@ interface Promise
...
      *     success, `$value` shall be `null` on failure.
```
enumag:
```suggestion
     * @param callable(\Throwable|null, TValue|null) $onResolved The first argument shall
     *     be `null` on success, while the second shall be `null` on failure.
```

enumag:
@kelunik Better?

kelunik:
@enumag Better, but not yet there, see https://psalm.dev/r/ca275ffaf2.

Adding [`@psalm-immutable`](https://psalm.dev/docs/annotating_code/templated_annotations/) to `Amp\Promise` helps: https://psalm.dev/r/0c9182b7b8

kelunik:
Adding `@psalm-mutation-free` to `onResolved` helps, too: https://psalm.dev/r/e01af1ee64

kelunik:
Thanks!

Issue #305, opened 2020-03-28
How to document Promise types / generics?


dantleech:
Hey,

Are there any plans / would it be possible to support phpstan/psalm generics for `Promise`?

```
class Foo
{
    /**
     * @return Promise<MyDomainThing>
     */
    public function wait(): Promise { // ... }
}
```

enumag:
See https://github.com/amphp/amp/pull/301

enumag:
But yeah we might want to apply the rest of the Psalm stubs too.

tncrazvan:
It's been 2 years, but just to be clear you're trying to achieve something like this right? (using psalm)
![image](https://user-images.githubusercontent.com/6891346/173188424-e29f0d34-447f-42b3-ab12-813260d7d482.png)

Psalm does it out of the box for me in VSCode, though you may want to disable autocompletion for [Intelephense](https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client) if you're using that plugin, it seems to conflict with psalm when it comes to PHPDocs ontop of duplicating some other autocompletions aswell.

Here's my psalm.xml, I hope it helps:

```xml
<?xml version="1.0"?>
<psalm
    errorLevel="7"
    resolveFromConfigFile="true"
    ensureArrayStringOffsetsExist="true"
    ensureArrayIntOffsetsExist="true"
    checkForThrowsInGlobalScope="true"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="https://getpsalm.org/schema/config"
    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
>
    <projectFiles>
        <directory name="src" />
        <directory name="tests" />
        <ignoreFiles>
            <directory name="vendor" />
        </ignoreFiles>
    </projectFiles>
</psalm>
```

ps: I can't seem to get it working in PHPStorm (with psalm) unfortunately.

PR #306, opened 2020-03-28
Add more generic phpdocs

dantleech:
See #305 

This allows:

```php
interface WatcherProcess
{
    public function stop(): void;

    /**
     * @return Promise<?ModifiedFile>
     */
    public function wait(): Promise;
}
```

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -342,9 +342,7 @@ namespace Amp\Promise
      * promise succeeds with an array of values used to succeed each contained promise, with keys corresponding to
      * the array of promises.
      *
-     * @param Promise[]|ReactPromise[] $promises Array of only promises.
-     *
-     * @psalm-param array<array-key, Promise|ReactPromise> $promises
+     * @param array<array-key, Promise|ReactPromise> $promises
      *
      * @return Promise
      *
@@ -740,12 +738,9 @@ namespace Amp\Iterator
      *
      * @template TValue
      *
-     * @param Iterator $iterator
+     * @param Iterator<TValue> $iterator
      *
-     * @psalm-param Iterator<TValue> $iterator
-     *
-     * @return Promise
-     * @psalm-return Promise<array<array-key, TValue>>
+     * @return Promise<array<array-key, TValue>>
      */
     function toArray(Iterator $iterator): Promise
     {


lib/Promise.php
```diff
@@ -23,7 +24,7 @@ interface Promise
      * Note: You shouldn't implement this interface yourself. Instead, provide a method that returns a promise for the
      * operation you're implementing. Objects other than pure placeholders implementing it are a very bad idea.
      *
-     * @param callable(\Throwable|null, TValue|null) $onResolved The first argument shall
+     * @param callable(\Throwable|null, TValue|null): void $onResolved The first argument shall
```
dantleech: Phpstan wouldn't parse this without `:void` (which is in the Psalm stub)

dantleech:
> Please add @template for the call function too

I tried this, but at least with PHPStan it gives an error, and I'm not sure how to add the type information:

```php
   /**
     * @template TValue
     *
     * @param callable():(\Generator<mixed, mixed, mixed, TValue>|TValue) $gen
     *
     * @return Promise<TValue>
     */
    function call(callable $callback, ...$args): Promise
```

```
 ------ ---------------------------------------------------------------- 
  Line   SystemDetector/CommandDetector.php                                     
 ------ ---------------------------------------------------------------- 
  16     Unable to resolve the template type TValue in call to function  
         Amp\call                                                        
 ------ ----------------------------------------------------------------
```

Untitled

lib/functions.php
```diff
@@ -12,9 +12,11 @@
      *
      * Use this function to create a coroutine-aware callable for a promise-aware callback caller.
      *
-     * @param callable(mixed ...$args): mixed $callback
+     * @template TValue
      *
-     * @return callable(mixed ...$args): \Amp\Promise
+     * @param callable():\Generator<mixed, mixed, mixed, TValue> $callback
```
enumag: `$callback` is passed to `call()` which accepts many different types as return types from the callback, not just Generator.
See https://github.com/amphp/amp/blob/master/lib/functions.php#L65-L77

enumag:
> I tried this, but at least with PHPStan it gives an error, and I'm not sure how to add the type information:

I see... yeah that complicates things. Maybe we should report it to PHPStan then.

lib/functions.php
dantleech: I've updated to

```
     * @param callable(mixed ...$args): (\Generator<mixed, mixed, mixed, TValue>|\Amp\Promise<TValue>|mixed) $callback
```

If that makes sense.

enumag: It would be better to use `TValue` instead of the last mixed but I'm guessing that runs into the same issue you mentioned above, right?

enumag: Second mixed of the generator should possibly be `Promise<mixed>`. Afaik yielding anything else than a Promise would cause an error.

dantleech: Actually, that works. However not sure about the ReactPromise? Should we add that too?

```
     * @param callable(mixed ...$args): (\Generator<mixed, mixed, mixed, TValue>|\Amp\Promise<TValue>|\React\Promise\PromiseInterface<TValue>|TValue) $callback
```

enumag: I guess... Personally I'm kinda against using React promises in Amphp code and would always call `adapt()` anyway but it works so I guess it's correct to use it.

By the way. technically the return value of the Generator is `Promise<TValue>|ReactPromise<TValue>|TValue`.

dantleech: cool, updated to use local names

lib/Promise.php
```diff
@@ -5,6 +5,7 @@
 /**
  * Representation of the future value of an asynchronous operation.
  *
+ * @template TValue
```
kelunik: Why do we need both?

enumag:
Pretty sure we don't.

lib/Promise.php
dantleech: Yep, phpstan seems happy with just `template-covariant` (i don't know what the difference is)

kelunik:
I've added quite a lot of annotations I had locally already, so there are some conflicts now, sorry about that.

lib/functions.php
```diff
@@ -14,8 +14,7 @@
      *
      * @template TReturn
      *
-     * @param callable(mixed ...$args):(\Generator<mixed,Promise|ReactPromise|array<array-key, Promise|ReactPromise>,mixed,Promise<TReturn>|ReactPromise|TReturn>|Promise<TReturn>|ReactPromise|TReturn)
-     *     $callback
+     * @param callable(mixed ...$args):(\Generator<mixed,Promise|ReactPromise|array<array-key, Promise|ReactPromise>,mixed,Promise<TReturn>|ReactPromise|TReturn>|Promise<TReturn>|ReactPromise|TReturn) $callback
```
dantleech: The phpstan docblock parser didn't like `$callback` being ona new line for some reason.

dantleech:
Updated most of it seems to be implemented now anyway, although it does fix a parsing issue with Phpstan. The new `call / asyncCall` annotations don't work with Phpstan for the reason given above.

lib/functions.php
kelunik: Could you please open an phpstan issue for that? Phpstorm automatically formats it like that for me.

dantleech: Hmm, it works now it seems.

dantleech:
ok - so I think everything is covered by your changes and the original stuff in this PR can be discarded.

I've updated the PR to replace `@psalm-param` with `@param` as these seem to work with Phpstan.

lib/functions.php
ondrejmirtes: PhpStorm doesn't understand this syntax at all so I'm not surprised it breaks on that.

ondrejmirtes:
I wouldn't use these in `@param` because you'll break IDE autocompletion. Better to use analyser-prefixed tags like `@phpstan-param` and `@psalm-param`.

dantleech:
Then, I'll close this. Thanks for implementing the types @kelunik 

@ondrejmirtes I would create an issue for what seems like a Phpstan bug above, but I'm not quite sure yet how to reproduce, hopefully can do it later.

dantleech:
I just tried running Psalm with the lastest annotations in master:

```
            while (null !== $chunk = yield $stream->read()) {
```
gives:

```
INFO: InvalidArgument - src/Parser/LineParser.php:18:36 - Argument 1 of str_split expects string, Amp\Promise<null
|string> provided (see https://psalm.dev/004)                                                                     
                foreach (str_split($chunk) as $char) {                                                             
```

(after adding the psalm-return to `ProcessInputStream`:

```
     * @psalm-return Promise<string|null>
     */
    public function read(): Promise
```

:thinking:

enumag:
Even with the annotations both Psalm and PHPStan are very unlikely to recognize that `yield Promise<TValue>` will produce `TValue`.

ondrejmirtes:
I havent checked the actual annotations but Generator stubs in PHPStan/Psalm support this in TSend.

lib/functions.php
kelunik: @ondrejmirtes: PhpStorm doesn't break, it just reformats too long lines like that.

kelunik:
@ondrejmirtes They just support one type per generator, but not a type based on the yielded value.

ondrejmirtes:
@kelunik I don't understand. See this example: https://phpstan.org/r/0eaa8d12-2237-489a-9165-eb243d637329

What do you think wouldn't work in regards to Amp?

lib/functions.php
ondrejmirtes: @kelunik Yeah but it won't be able to understand the method accepts `callable`. Maybe in this case yes, because the type starts with `callable`, but definitely not in some more complex cases.

enumag:
@ondrejmirtes With Amphp the Generators are usually immediately called using the `call()` function. They do not have an annotation with a TSend type. Also they often yield promises for different things. See the code example [here](https://amphp.org/amp/coroutines/).

`$http->request("https://example.com/")` retuns a `Promise<Response>` which means that `$response instanceof Response`. Then `$response->getBody();` returns a `Promise<string>` so `$body` is a string.

enumag:
In other words this code:

```php
function getExampleCom(HttpClient $http) {
    return Amp\call(function () use ($http) {
        $response = yield $http->request("https://example.com/");
        $body = yield $response->getBody();
        return $body;
    });
}
```

would look like this with async-await:

```
async function getExampleCom(HttpClient $http) {
    $response = await $http->request("https://example.com/");
    $body = await $response->getBody();
    return $body;
}
```

dantleech:
The issue seems similar to: https://github.com/phpstan/phpstan-doctrine/issues/101

PR #308, opened 2020-04-14
Add Loop::isRunning()

trowski:
`Amp\Promise\wait()` now fails if called within a running event loop, which has always been a mistake anyway.

### DIFF
--- a/lib/Loop.php
+++ b/lib/Loop.php
@@ -109,6 +109,16 @@ final class Loop
     }
 
     /**
+     * Returns true if the event loop is running, false if it is stopped.
+     *
+     * @return bool True if the event loop is running, false if it is stopped.
+     */
+    public static function isRunning(): bool
+    {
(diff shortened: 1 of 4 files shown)


trowski:
Tested with `amphp/parallel`. All tests except one using `wait()` within an `AsyncTestCase` test pass.

trowski:
> One approach was `Loop::execute()` setting the running flag to its previous value [here](https://github.com/amphp/amp/blob/2053b42fdbf522c6bce1e786f08e554a92ae0368/lib/Loop/Driver.php#L75).

I was never thrilled with the current behavior, it seems unintuitive. Maybe we should change it: #309.

> I'm not sure whether we should emit a warning or throw an exception.

If we go with this approach, I'd lean toward throwing the exception.

trowski:
Closing in favor of #311.

Issue #84, opened 2017-03-13
Stopping nested loops


trowski:
Nested loops can result in cases where multiple calls to stop are timed in such a way that the lower-level loop does not stop.

This code continues the base loop indefinitely, even though it seems it should be stopped.

```php
Loop::run(function () {
    Loop::repeat(1000, function () {
        print microtime(true) . PHP_EOL;
    });

    Loop::defer(function () {
        Loop::stop();
    });

    Loop::run(function () {
        Loop::stop();
    });
});
```

While this is an edge-case, I can see this potentially happening if multiple calls to `wait()` are made and the promises are resolved in the same tick.

This happens because we ignore multiple calls to stop. Perhaps multiple calls should not be ignored, but rather stop the prior loop if running as a nested loop.

kelunik:
I think we shouldn't care about this case and instead forbid `Amp\wait` in case the loop is running.

But that might not work for shutdown functions, right?

trowski:
We'd need a way to tell if the loop is running to forbid it in `Amp\wait`. Not a problem to add, but currently is missing from `Driver`.

This could be an issue in shutdown functions as you suggested. If each function starts their own event loop, subsequently calling `stop`, the likelihood of calls to `stop` occurring in the same tick might be may be high…

kelunik:
@trowski We already have a `"running"` boolean in `getInfo()`.

kelunik:
Would be cool if we could access https://github.com/php/php-src/blob/c6982995504b6e21e8a5ade29cfb16a55196dc43/main/main.c#L1808 and disable the recursive run protection in that case.

kelunik:
@trowski I think a single `Loop::stop()` should just stop all recursive runs and further `Loop::stop()` calls should still be ignored.

trowski:
Addressed with https://github.com/amphp/amp/commit/173549ebbac46c427627a11e07287b938a063d41.

PR #309, opened 2020-04-14
Stop nested loops individually

trowski:
Related to #308.

This requires `Loop::stop()` to be called for each level of loop nesting, which I think is more intuitive than the current behavior. I'm not sure if we documented that all levels of loop nesting are stopped, so I might consider the current behavior as undefined. This of course matters only if the same `Driver` instance is used when nested. Using two instances and calling stop on only one will not stop the other. With this PR, a single instance behaves the same as two instances.

With this behavior, calling `Amp\Promise\wait()` within `Loop::run()` doesn't matter, as only the inner loop will be stopped.

### DIFF
--- a/lib/Loop/Driver.php
+++ b/lib/Loop/Driver.php
@@ -77,6 +77,14 @@ abstract class Driver
     }
 
     /**
+     * @return bool True if the event loop is running, false if it is stopped.
+     */
+    public function isRunning(): bool
+    {
+        return $this->running;
+    }
+
+    /**
      * @return bool True if no enabled and referenced watchers remain in the loop.
      */
     private function isEmpty(): bool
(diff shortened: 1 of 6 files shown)


kelunik:
@bwoebi What's your opinion here?

kelunik:
Related to #84.

lib/Loop/Driver.php
```diff
@@ -62,20 +62,28 @@ abstract class Driver
      */
     public function run()
     {
-        $this->running = true;
+        $nesting = $this->nesting++;
 
         try {
-            while ($this->running) {
+            while ($nesting < $this->nesting) {
```
kelunik, line 68: Shouldn't this be `$this->isRunning()`?

trowski: No. We want this particular while loop to end if the nesting level goes below the value when it was started. `$this->isRunning()` is true regardless of nesting depth.

PR #311, opened 2020-04-15
Allow nested calls to Amp\Promise\wait

kelunik:
This is an improved implementation for #309 that doesn't suffer from the same problems and has improved BC, because normal nested `Loop::run()` calls are unaffected. The only change needed that's slightly breaking is if a custom driver was implemented that delegates all methods like `TracingDriver`.

This implementation could also be used by Symfony's HTTP client, which does have [something similar to `Amp\Promise\wait`](https://github.com/symfony/http-client/blob/3846a7fceb2c7e6c05f8edd1bd999a2f6be89623/Response/AmpResponse.php#L167-L181).

We could change the callable to a proper object to improve typing.

### DIFF
--- a/lib/Loop/DelegateLoop.php
+++ b/lib/Loop/DelegateLoop.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Amp\Loop;
+
+final class DelegateLoop
+{
+    private $run;
+    private $stop;
+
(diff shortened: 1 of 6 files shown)


kelunik:
Closing in favor of the upcoming v3.

PR #313, opened 2020-04-20
Allow nested calls to Amp\Promise\wait

kelunik:
Compared to #311, this doesn't break BC at all, because we previously threw an `\Error` in this case, which indicates a programming error and not valid runtime behavior.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -214,16 +214,17 @@ namespace Amp\Promise
         $resolved = false;
 
         try {
-            do {
-                Loop::run(function () use (&$resolved, &$value, &$exception, $promise) {
-                    $promise->onResolve(function ($e, $v) use (&$resolved, &$value, &$exception) {
-                        Loop::stop();
-                        $resolved = true;
-                        $exception = $e;
-                        $value = $v;
-                    });
-                });
-            } while (!$resolved && Loop::getInfo()['enabled_watchers']['referenced'] > 0);
+            $promise->onResolve(function ($e, $v) use (&$resolved, &$value, &$exception) {
+                Loop::stop();
+
+                $resolved = true;
+                $exception = $e;
+                $value = $v;
+            });
+
+            while (!$resolved && Loop::getInfo()['enabled_watchers']['referenced'] > 0) {
+                Loop::run();
+            }
         } catch (\Throwable $throwable) {
             throw new \Error("Loop exceptionally stopped without resolving the promise", 0, $throwable);
         }
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -212,15 +212,20 @@ namespace Amp\Promise
         }
 
         $resolved = false;
+        $running = false;
 
         try {
-            $promise->onResolve(function ($e, $v) use (&$resolved, &$value, &$exception) {
-                Loop::stop();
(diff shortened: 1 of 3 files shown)


enumag:
Personally I prefer an error to be honest. It wasn't something I did intentionally.

kelunik:
@enumag Calling `Amp\Promise\wait` inside `Loop::run()` is a different but related topic and isn't solved by this PR, as this PR will still stop the outer `Loop::run()`.

trowski:
> @enumag Calling `Amp\Promise\wait` inside `Loop::run()` is a different but related topic and isn't solved by this PR, as this PR will still stop the outer `Loop::run()`.

Which I would also like to see either work in some regard or error. I'm starting to lean back toward the simple solution in #309 where both nested calls and calls within `Loop::run()` fail.

lib/functions.php
```diff
@@ -214,14 +214,16 @@ function wait($promise)
...
+                Loop::run(function () use (&$resolved, &$value, &$exception, $promise) {
+                    $promise->onResolve(function ($e, $v) use (&$resolved, &$value, &$exception) {
+                        Loop::stop();
+                        $resolved = true;
+                        $exception = $e;
+                        $value = $v;
+                    });
                 });
-            });
+            } while (!$resolved && Loop::getInfo()['enabled_watchers']['referenced'] > 0);
```
trowski, line 226: Since the promise is resolved, `Loop::stop()` will be called on every tick of the event loop.

kelunik: Fixed.

lib/functions.php
```diff
@@ -214,16 +214,17 @@ function wait($promise)
...
-            } while (!$resolved && Loop::getInfo()['enabled_watchers']['referenced'] > 0);
+            $promise->onResolve(function ($e, $v) use (&$resolved, &$value, &$exception) {
+                Loop::stop();
+
+                $resolved = true;
+                $exception = $e;
+                $value = $v;
+            });
+
+            while (!$resolved && Loop::getInfo()['enabled_watchers']['referenced'] > 0) {
```
trowski, line 225: Should we make `Driver::isEmpty()` public so this doesn't feel as hacky?

kelunik: This would be a slight breaking change for drivers like `TracingDriver`, while the current implementation uses public API.

psalm-shepherd:
Psalm found errors in other files

```
lib/functions.php:237:14 - Found a redundant condition when evaluating $resolved of type empty-mixed and trying to reconcile it with a falsy assertion
```

psalm-shepherd:
Psalm found errors in other files

```
lib/functions.php:237:14 - Found a redundant condition when evaluating $resolved of type empty-mixed and trying to reconcile it with a falsy assertion
```

psalm-shepherd:
Psalm found errors in other files

```
lib/functions.php:237:14 - Found a redundant condition when evaluating $resolved of type empty-mixed and trying to reconcile it with a falsy assertion
```

psalm-shepherd:
Psalm found errors in other files

```
lib/functions.php:237:14 - Found a redundant condition when evaluating $resolved of type empty-mixed and trying to reconcile it with a falsy assertion
```

kelunik:
Closing in favor of the upcoming v3.

PR #315, opened 2020-05-04
Add Amp\Iterator\discard

kelunik:
Relates to https://github.com/amphp/sync/pull/15 to simply consume the resulting iterator.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -763,6 +763,31 @@ namespace Amp\Iterator
     }
 
     /**
+     * Discards all remaining items and returns the number of discarded items.
+     *
+     * @template TValue
+     *
+     * @param Iterator $iterator
+     *
+     * @return Promise
+     *
+     * @psalm-param Iterator<TValue> $iterator
+     * @psalm-return Promise<int>
+     */
+    function discard(Iterator $iterator): Promise
+    {
+        return call(static function () use ($iterator) {
+            $count = 0;
+
+            while (yield $iterator->advance()) {
+                $count++;
+            }
+
+            return $count;
+        });
+    }
+
+    /**
      * Collects all items from an iterator into an array.
      *
      * @template TValue
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -776,7 +776,7 @@ namespace Amp\Iterator
      */
     function discard(Iterator $iterator): Promise
     {
-        return call(static function () use ($iterator) {
+        return call(static function () use ($iterator): \Generator {
             $count = 0;
 
             while (yield $iterator->advance()) {
--- a/test/IteratorDiscardTest.php
+++ b/test/IteratorDiscardTest.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Amp\Test;
+
+use Amp\Iterator;
+use Amp\PHPUnit\AsyncTestCase;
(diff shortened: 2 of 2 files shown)


lib/functions.php
```diff
@@ -762,6 +762,31 @@ function concat(array $iterators): Iterator
...
+     * @param Iterator $iterator
+     *
+     * @return Promise
+     *
+     * @psalm-param Iterator<TValue> $iterator
+     * @psalm-return Promise<int>
+     */
+    function discard(Iterator $iterator): Promise
+    {
+        return call(static function () use ($iterator) {
```
enumag, line 779:
```suggestion
        return call(static function () use ($iterator): \Generator {
```

test/IteratorDiscardTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Amp\Test;
+
+use Amp\Iterator;
+use Amp\PHPUnit\AsyncTestCase;
+use function Amp\Iterator\discard;
+
+class IteratorDiscardTest extends AsyncTestCase
+{
+    public function testEmpty()
```
enumag, line 11:
```suggestion
    public function testEmpty(): \Generator
```

test/IteratorDiscardTest.php
```diff
@@ -0,0 +1,20 @@
...
+    {
+        $this->assertSame(0, yield discard(Iterator\fromIterable([])));
+    }
+
+    public function testCount()
```
enumag, line 16:
```suggestion
    public function testCount(): \Generator
```

lib/functions.php
```diff
@@ -762,6 +762,31 @@ function concat(array $iterators): Iterator
...
+            $count = 0;
+
+            while (yield $iterator->advance()) {
+                $count++;
```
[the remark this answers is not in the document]
kelunik: Based on what?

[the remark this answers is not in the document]
enumag: Some benchmarks I saw a long time ago. Also it makes sense since postincrementation needs to preserve the original value for the expression where it's used.

lib/functions.php
```diff
@@ -762,6 +762,31 @@ function concat(array $iterators): Iterator
...
+     * @template TValue
+     *
+     * @param Iterator $iterator
+     *
+     * @return Promise
+     *
+     * @psalm-param Iterator<TValue> $iterator
+     * @psalm-return Promise<int>
+     */
+    function discard(Iterator $iterator): Promise
```
kelunik, line 777: I'd name it `count`, but I guess it might be misused and people will complain that the iterator can't be reused then.

trowski:
A thought – a function returning an `Iterator` that emits without yielding to the emit promise would do the same thing, but potentially be more useful than just for testing. `removeBackpressure()` or some other name? 
@kelunik Think that would be better?

kelunik:
@trowski That'd buffer the entire iterator in memory.

trowski:
Oh yeah… bah, so you'd still have to iterate it anyway, defeating the purpose.

lib/functions.php
kelunik: @trowski Should it return `Promise<void>` instead of should we keep the count?

trowski: Keep the count.

PR #316, opened 2020-05-11
Add Streams

trowski:
`Stream` is a replacement for `Iterator`, merging iteration into a single method and fixing some of the minor issues we've discovered with iterators after using them for the past couple of years.

- `Stream::continue()` returns a promise that is fulfilled with the emitted value or `null` if the stream has completed. The promise may also fail with the same exception used to fail the stream.
- Steams cannot emit promises. Yielding promises is behaviorally strange (iteration order now depends on resolution order) and also generally unnecessary.
- Streams cannot emit `null` as it is used to signal the end of the stream.
- `Stream::continue()` may be called an unlimited number of times. The promises returned are resolved as values are emitted. If the stream ends with fewer values, excess promises are resolved with `null`.
- `StreamSource` is analogous to `Iterator`'s `Emitter`. It is used to emit values and complete or fail the stream. Like `Emitter`, it is not meant to be part of a public API.
- `AsyncGenerator` is analogous to `Iterator`'s `Producer`. `AsyncGenerator` implements a full async-generator, including `send()` and `throw()` methods like `\Generator`. The promise returned from the `$emit` callable is resolved with `null` if `continue()` is used, otherwise it is resolved with the sent value or failed with the thrown exception. `yield` is used as `await` like any other coroutine. A callable `$emit` is provided as the single argument to function provided to the `AsyncGenerator` constructor for emitting values. Using send/throw is not required of course, and can be used like any other `Stream` as `\Generator` can be used like an `\Iterator`.
- The promise returned from `StreamSource::emit()` and the emit callable provided to `AsyncGenerator` resolves once the value is consumed from the stream or fails with `DisposedException` if the stream is destroyed.

- When a `Stream` is destroyed, the stream source is notified that the stream is no longer being consumed through a `DisposedException` failing the promise returned from `StreamSource::emit()` (or the emit callable provided to `AsyncGenerator`).
- Only a single `Stream` object may be associated with a source. Since iteration should only be occurring in one place, this just makes sense and makes disposal simple.
- All back-pressure is removed if a stream is destroyed (disposed).


lib/Internal/Yielder.php
```diff
@@ -120,25 +130,28 @@ private function yield($value): Promise
...
-
         if ($value instanceof Promise || $value instanceof ReactPromise) {
             throw new \TypeError("Streams cannot yield promises");
         }
 
+        if ($key === null) {
+            $key = $this->nextKey++;
+        } elseif (\is_int($key) && $key >= $this->nextKey) {
+            $this->nextKey = $key + 1;
+        }
```
kelunik, line 141: IMO we should only allow all keys being `null` or none, as a mixture is very likely to cause bugs.

lib/AsyncGenerator.php
```diff
@@ -0,0 +1,63 @@
+<?php
+
+namespace Amp;
+
+/**
+ * @template TValue
+ * @template TKey
+ */
+final class AsyncGenerator implements Stream
```
enumag, line 9: I really dislike the name of this class. From this name I expected something like normal PHP's Generator except Asynchronous. Meaning it should be able to both emit values AND receive them using something similar to `Generator::send();`:

```
$recievedValue = yield $yield($emittedValue);
```

Since this isn't the case I highly recommend renaming this class to something more appropriate.

Of course this kind of stream would have no backpressure since it needs to receive a value before it can continue execution. I'm not necessarily suggesting for you to add it since the use case for it is rare, but the name here is very confusing for me.

lib/Internal/Yielder.php
trowski: Do you mean just auto-generate all keys?

lib/AsyncGenerator.php
trowski: Very much open to name suggestions. I agree that we should keep `AsyncGenerator` as the name of a class that does what you describe, in case a use does arise.

lib/Internal/Yielder.php
kelunik: I mean either all yields must provide a non-null key or every yield must pass `null` per stream.

trowski: Or do you rather mean once a custom key is used, error if a custom key is not provided. That definitely makes sense.

kelunik: Now that I think about it more, these should probably be two different data structures. It's like a list vs. map. If I filter a list, I expect the indices to be without gaps.

lib/AsyncGenerator.php
enumag: I admit that naming this class is a bit tough. I do like the name Producer which is currently used for async iterators. So possibly `StreamProducer` to make it clear it's related to the new streams feature?

lib/Internal/Yielder.php
trowski: I honestly don't think keys are very useful here, which is why I didn't include them in the first place. I think forbidding `null` made more sense.

trowski: Updated to only auto-generate keys.

lib/AsyncGenerator.php
trowski: Naming is hard and I wanted to try doing it, so I actually implemented it as an async generator. 🤓

enumag: @trowski 🥇

trowski: I seem to have an ordering issue to work out… guess it wasn't as simple as I hoped.

trowski: Now it should work as expected! 😁

kelunik: I don't think we need `AsyncGenerator`, I'm sure we can find a better name for it.

trowski: Why not? Maybe there's something interesting that could be implemented with it. And it still works the same as `Producer` if you're not interested in send/throw.

trowski:
This is now ready for another review. I look forward to using `AsyncGenerator`. I believe we can create some powerful generator parsers and async iterators.

lib/Internal/DisposableStream.php
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Wraps a Stream instance that has public methods to yield, complete, and fail into an object that only allows
+ * access to the public API methods and sets $disposed to true when the object is destroyed.
+ *
+ * @internal
+ *
+ * @template-covariant TValue
+ * @template-implements Stream<TValue>
+ */
+class DisposableStream implements Stream
```
kelunik, line 17: I'd expect a `dispose()` method on a `DisposableStream`

bwoebi: Maybe call it `DestructibleStream` then?

lib/DisposedException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Amp;
+
+/**
+ * Will be thrown from {@see StreamSource::yield()} or the yield callable provided by {@see AsyncGenerator} if the
+ * associated stream is destroyed.
+ */
+final class DisposedException extends \Exception
```
bwoebi, line 9: `AlreadyDestroyedException` sounds more self-explaining

lib/Internal/DisposableStream.php
kelunik: No, I'd offer a `dispose()` method, it can be useful if you have a circular deference and don't want to call `gc_collect_cycles()`.

lib/DisposedException.php
kelunik: It's not realy destroyed if we offer a `dispose()` method.

lib/Internal/DisposableStream.php
trowski: Added.

lib/TransformationStream.php
```diff
@@ -26,9 +30,24 @@ public function dispose()
...
+        }
+
+        return new self($this->apply($operator));
+    }
+
+    private function apply(callable $operator): self
+    {
+        $stream = $operator($this);
+
+        if ($stream instanceof Stream) {
```
kelunik, line 46:
```suggestion
        if (!$stream instanceof Stream) {
```

kelunik:
I think I'd like to see failing only intermediate values instead of the whole stream. I don't have an API suggestion for that, yet, but just want to write it down here so we don't forget about it.

lib/Internal/EmitSource.php
```diff
@@ -0,0 +1,302 @@
...
+    private $used = false;
+
+    /**
+     * @return Promise<mixed|null>
+     *
+     * @psalm-return Promise<TValue|null>
+     */
+    public function continue(): Promise
+    {
+        return $this->next(new Success);
```
kelunik, line 64: We're creating a lot of unnecessary objects here, maybe store the `new Success` in a static field, as it's basically immutable?

lib/Internal/EmitSource.php
```diff
@@ -0,0 +1,302 @@
...
+        $this->completed = !$disposed;
+        $this->disposed = $disposed;
+
+        if ($this->result) {
+            return;
+        }
+
+        \assert((function () {
+            $env = \getenv("AMP_DEBUG") ?: "0";
+            if (($env !== "0" && $env !== "false") || (\defined("AMP_DEBUG") && \AMP_DEBUG)) {
```
kelunik, line 275: We should probably turn this into a function like `isDebugEnabled()`.

lib/functions.php
```diff
@@ -812,4 +810,300 @@ function toArray(Iterator $iterator): Promise
...
+    ): Stream {
+        if (!$iterable instanceof \Traversable && !\is_array($iterable)) {
+            throw createTypeError(["array", "Traversable"], $iterable);
+        }
+
+        if ($delay) {
+            return new AsyncGenerator(static function (callable $yield) use ($iterable, $delay) {
+                foreach ($iterable as $value) {
+                    yield new Delayed($delay);
+                    yield $yield($value instanceof Promise ? yield $value : $value);
```
kelunik, line 876: Needs to support ReactPromise as well.

lib/functions.php
```diff
@@ -812,4 +810,300 @@ function toArray(Iterator $iterator): Promise
...
+                }
+            });
+        }
+
+        return new AsyncGenerator(static function (callable $yield) use ($iterable) {
+            foreach ($iterable as $value) {
+                yield $yield($value instanceof Promise ? yield $value : $value);
```
kelunik, line 883: Needs to support ReactPromise as well.

lib/functions.php
```diff
@@ -812,4 +810,300 @@ function toArray(Iterator $iterator): Promise
...
+        $promise->onResolve(static function ($exception) use ($source) {
+            if ($exception) {
+                $source->fail($exception);
+                return;
+            }
+
+            $source->complete();
+        });
+
+        return $source->stream();
```
kelunik, line 1034: This implementation feels overly complicated and is probably much easier to implement using `AsyncGenerator` and a nested loop over all streams.

lib/Internal/EmitSource.php
trowski: Yeah, I could definitely optimize that away.

trowski:
> I think I'd like to see failing only intermediate values instead of the whole stream. I don't have an API suggestion for that, yet, but just want to write it down here so we don't forget about it.

I think it makes more sense to keep it simple. If a user wants to represent failed values, they can emit exceptions.

`AsyncGenerator::throw()` could also be considered a way to emit a failed value, but that of course is built on top of the basic stream implementation.

lib/functions.php
trowski: I copied/pasted from the `Iterator` implementation. Wow, I have no idea why the implementation was so complex. I'll change the one for `Iterator` too.

lib/Internal/AutoDisposingPipeline.php
```diff
@@ -26,7 +26,7 @@ public function __construct(EmitSource $source)
 
     public function __destruct()
     {
-        $this->source->dispose();
+        $this->source->destroy();
```
kelunik, line 29: Why destroy instead of dispose?

trowski: The generator can keep running even if the `AsyncGenerator` object was destroyed, so it can keep emitting values to pending promises from `continue()` calls before the instance was GC'd.

trowski:
This work eventually lead to the [pipeline library](https://github.com/amphp/pipeline) for v3.

PR #319, opened 2020-05-25
Enforce timer interval as minimum time to execution

kelunik:
Relates to https://github.com/symfony/symfony/pull/36960.

### DIFF
--- a/test/Loop/DriverTest.php
+++ b/test/Loop/DriverTest.php
@@ -662,9 +662,9 @@ abstract class DriverTest extends TestCase
                                 // explicitly use *different* streams with *different* resource ids
                                 $ends = \stream_socket_pair(
                                     \stripos(
-                                    PHP_OS,
-                                    "win"
-                                ) === 0 ? STREAM_PF_INET : STREAM_PF_UNIX,
+                                        PHP_OS,
+                                        "win"
+                                    ) === 0 ? STREAM_PF_INET : STREAM_PF_UNIX,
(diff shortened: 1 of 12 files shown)


test/Loop/DriverTest.php
```diff
@@ -106,6 +107,52 @@ public function testLoopRunsCanBeConsecutiveAndNested()
...
+
+                \usleep(500000);
+            });
+        });
+
+        $this->assertNotSame(0, $start);
+        $this->assertNotSame(0, $invoked);
+
+        $this->assertGreaterThan(1000, $invoked - $start);
+        $this->assertLessThan(1100, $invoked - $start);
```
kelunik, line 131: This is currently expected to fail, because watchers are only activated in the next tick.

nicolas-grekas:
I'm running into this issue on https://github.com/symfony/symfony/pull/37136.
Here is my local reproducer:
`./phpunit src/Symfony/Component/HttpClient/ --filter 'Amp.*(Vulcain|Pause)'`
Any idea about what could be causing this? Any workaround maybe?

nicolas-grekas:
I had a closer look and this relates to the `$nowUpdateNeeded` property: forcing it to `true` by wrapping the `Loop::delay()` inside `Loop::defer()` work around the issue. The reason is that a `tick()` happens before the call.

The fix could be to set `$nowUpdateNeeded` to `true` when `Loop::delay()` is called.

Issue #265, opened 2019-03-03
Stop suppressing stream_select errors


ostrolucky:
I spent some hours today debugging why do my loops hang. There are numerous occasions when `stream_select` does not work. Instead of throwing the error, loop is going forever, because for some unknown reason, [you throw only for one single type of error](https://github.com/amphp/amp/blob/272f78dcb2897bead923fd5f94c9b88b18e40f70/lib/Loop/NativeDriver.php#L179). 

Example 1:

```php
Loop::run(function() {
    $stream = fopen('php://memory', 'rwb');

    Loop::onWritable($stream, function() {
        throw new \Exception('I should have been thrown');
    });

    fwrite($stream, 'wat');
});
```

Example 2:

```php
Loop::run(function() {
    $stream = tmpfile();
    stream_filter_append($stream, "string.rot13", STREAM_FILTER_WRITE);

    Loop::onWritable($stream, function() {
        throw new \Exception('I should have been thrown');
    });

    fwrite($stream, 'wat');
});
```

Hereby I request you to stop ignoring stream_select errors as it leads to hard to debug issues.

trowski:
I'm not entirely sure why only a single type of error is being thrown if `stream_select()` returned `false`. @kelunik Do you remember why this was done?

@ostrolucky Note that the streams you are passing to the loop in the examples above cannot be used with the event loop. Only socket can be polled. Memory streams and file streams will never invoke the watcher, thus the exceptions in your watchers can never be thrown. The error suppression was stopping you from seeing the error from `stream_select()` that the streams were invalid, not catching any exceptions.

ostrolucky:
Yes I know now, see
> There are numerous occasions when stream_select does not work

I'm not blaming you, that's stream_select limitation. But I didn't that when I was trying it. I would realize that much sooner if error wasn't silenced :)

trowski:
I guess `'I should have been thrown'` as the exception message made me think you were expecting that to be thrown but your complaint was that it was not. We're on the same page now, no worries.

ostrolucky:
Actually rereading your comment, this is not entirely true

> Memory streams and file streams will never invoke the watcher

Second example will work when removing stream_filter_append. That's what's causing stream_select to not work. File resources are ok.

trowski:
File resources appear to work, but will always report as readable or writable. So no, they don't actually work.

ostrolucky:
Yep, except those with stream_filter :)

ostrolucky:
[...]

PR #322, opened 2020-07-09
Throw from all stream_select() errors

hussein97-k:
Related to #265.

                $this->error(new \RuntimeException(\error_get_last()["message"]));


kelunik:
Closing in favor of #338. We can't throw for all errors, because signal interruptions are expected to throw a warning.

Issue #329, opened 2020-10-23
UvDriver Signal Example Issue/Suggestion (100% cpu)


duckboy81:
I'm back with another one!

Was attempting to implement the signal handling demo [examples/event-loop/signals.php](https://github.com/amphp/amp/blob/master/examples/event-loop/signals.php) and kept running into an issue where my script would catch the signal, hit `exit(0)` and hang at 100% cpu usage.

Switching from `uv` to the native driver seemed to work. But I was looking at a way to still use `uv` and it looks like swapping the `exit` call for a `Loop::stop()` call works well. The downside being that code execution stops at an arbitrary point after the `Loop::run()` line -- probably okay for the demo, but on a larger scale could be hard to manage.

I'll make a PR for it in a minute here.

php7.4/Linux

duckboy81:
Closing. Reopen if others experience the same.

PR #330, opened 2020-10-23
Allow signal handling for Uv w/o 100% cpu usage

duckboy81:
Ref #329

To be honest this could lend itself to its own `Uv` specific example file, but here's something that could work regardless of driver used.

Again, downside is that the end of code execution is not in the signal handling function, but sometime after the `Loop::run()` line.

### DIFF
--- a/examples/event-loop/signals.php
+++ b/examples/event-loop/signals.php
@@ -7,18 +7,10 @@ use Amp\Loop;
 
 print "Press Ctrl+C to exit..." . PHP_EOL;
 
-Loop::onSignal(SIGINT, function () {
+Loop::onSignal(SIGINT, function ($watcherId) {
     print "Caught SIGINT, exiting..." . PHP_EOL;
     
-    // Check for a Uv driver
-    if (Loop::get() instanceof Amp\Loop\UvDriver) {
-
-        // Stop the loop
-        Loop::stop();
-
-        // Cannot exit out of a UvDriver loop here, can only stop the loop
-        return;
-    }
+    Loop::cancel($watcherId);
     
     exit(0);
 });
--- a/examples/event-loop/signals.php
+++ b/examples/event-loop/signals.php
@@ -9,7 +9,20 @@ print "Press Ctrl+C to exit..." . PHP_EOL;
 
 Loop::onSignal(SIGINT, function () {
     print "Caught SIGINT, exiting..." . PHP_EOL;
+    
+    // Check for a Uv driver
+    if (Loop::get() instanceof Amp\Loop\UvDriver) {
+
+        // Stop the loop
+        Loop::stop();
+
+        // Cannot exit out of a UvDriver loop here, can only stop the loop
+        return;
+    }
+    
     exit(0);
 });
 
 Loop::run();
+
+exit(0);


enumag:
Not a very good solution imo. Now the signal handler can't control the exit code.

examples/event-loop/signals.php
```diff
@@ -9,7 +9,20 @@
...
+    // Check for a Uv driver
+    if (Loop::get() instanceof Amp\Loop\UvDriver) {
+
+        // Stop the loop
+        Loop::stop();
+
+        // Cannot exit out of a UvDriver loop here, can only stop the loop
+        return;
+    }
+
```
kelunik, line 22: What happens if you cancel the signal watcher before existing? Did you look where exactly it hangs in probably an infinite loop causing the 100% CPU usage?

```suggestion
Loop::onSignal(SIGINT, function ($watcherId) {
    print "Caught SIGINT, exiting..." . PHP_EOL;
    
    Loop::cancel($watcherId);
    
```

duckboy81: Not sure where my comment ended up, so sorry if this is a double tap.

Just tried this code and it still seems to get stuck in a loop. I tried debugging it for a couple hours, but I didn't get anywhere. Xdebug seemed to dump the profiler data before it got stuck in the loop, which made me think it was a `Uv` lib problem. (trying without xdebug loaded results in the same problem)

kelunik:
I can't reproduce 100% CPU usage, but I get a warning:

```
Press Ctrl+C to exit...
^CCaught SIGINT, exiting...
PHP Warning:  zend_signal: handler was replaced for signal (2) after startup in Unknown on line 0
```

duckboy81:
> I can't reproduce 100% CPU usage, but I get a warning:
> 
> ```
> Press Ctrl+C to exit...
> ^CCaught SIGINT, exiting...
> PHP Warning:  zend_signal: handler was replaced for signal (2) after startup in Unknown on line 0
> ```

Using `uv` as the driver? (I've currently got `uv` 0.2.4-3 if it matters)

kelunik:
Please try with the latest `master` of https://github.com/bwoebi/php-uv.

duckboy81:
> Please try with the latest `master` of https://github.com/bwoebi/php-uv.

Alright, give me a hot minute to try it.

PS: Thanks for being so active, you guys are awesome.

duckboy81:
> > Please try with the latest `master` of https://github.com/bwoebi/php-uv.
> 
> Alright, give me a hot minute to try it.
> 
> PS: Thanks for being so active, you guys are awesome.

No luck. Same result.

I'm not sure this is worth putting too much work into unless someone else runs into the same issue. They can reference this conversation.

kelunik:
Which PHP version do you use?

duckboy81:
Using the following:

> PHP 7.4.2 (cli) (built: Jan 21 2020 11:35:20) ( NTS )
> Zend Engine v3.4.0, Copyright (c) Zend Technologies
>     with Zend OPcache v7.4.2, Copyright (c), by Zend Technologies

If I switch from `uv` to the native driver, no issues too.

trowski:
Perhaps there's an issue in `ext-uv` when calling `exit` inside a watcher callback? Exit handling changed recently to throw an exception internally (not sure which PHP version this happened in).

Issue #331, opened 2020-11-01
Incorrect heap implementation in TimerQueue


descawed:
I've been using Amp's concurrent iterators to run many HTTP requests asynchronously with a small delay via `Amp\Iterator\fromIterable`. The delay is only supposed to be a few milliseconds, but I observed that every so often, there would be a long delay of a number of seconds before the next request started. I discovered this is due to an issue in `Amp\Loop\Internal\TimerQueue->removeAndRebuild`. This method doesn't account for the possibility that the node swapped in place of the removed node may need to be moved up the heap instead of down. This can result in timers that are expiring soon getting blocked behind timers that aren't expiring until later. I'll submit a PR shortly with a test illustrating this and a proposed solution.

PR #332, opened 2020-11-01
Correct heap implementation in TimerQueue

descawed:
Fix for #331.


kelunik:
Don't worry about codeclimate issues, it's not a required check to pass before merging. :-)

PR #339, opened 2021-01-10
Fix TypeError in EventDriver

rybakit:
Fixes

```
PHP Fatal error:  Uncaught TypeError: Cannot assign null to property Amp\Loop\EventDriver::$handle of type EventBase in .../lib/Loop/EventDriver.php:149
Stack trace:
#0 [internal function]: Amp\Loop\EventDriver->__destruct()
#1 {main}
  thrown in .../lib/Loop/EventDriver.php on line 149
```


kelunik:
Thanks!

PR #354, opened 2021-06-30
Fix for output buffering issue

shandak:
When several 'async' functions contain something like
```
ob_start();
....
Do some async stuff, like database select and 'echo' result
...
echo 'something';
$content = ob_get_contents();
ob_end_clean();
```
It's getting a mess with the result in `$content`

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -25,7 +25,7 @@ namespace Amp
             $promise = Promise\all($promise);
         }
 
-        $fiber = \Fiber::this();
+        $fiber = \Fiber::getCurrent();
         $resolved = false;
 
         if ($fiber) { // Awaiting from within a fiber.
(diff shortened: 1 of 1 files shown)


kelunik:
@shandak Output buffering is global, so concurrent output is expected to be "a mess". A better solution would be to write to some kind of `OutputStream` IMO instead of using `echo`.

shandak:
> @shandak Output buffering is global, so concurrent output is expected to be "a mess". A better solution would be to write to some kind of `OutputStream` IMO instead of using `echo`.

Agreed, I would not recommend using global output buffering too, but the goal here is to start using 'fibers' with already developed solutions without dramatic changes. For instance, in my case, it sounds almost impossible because `ob_start` is everywhere.

Maybe as another solution, it would be better to add event triggers around `Fiber::suspend` so everybody can add some custom behaviour for a specific purpose?

kelunik:
Supporting custom hooks for `Fiber::suspend` is likely a good idea, but we need to ensure it's lightweight enough, as it will happen quite frequently.

trowski:
@shandak This got closed unintentionally without comment when I was renaming some branches. Apparently the `v3` branch was temporarily deleted.

As amphp v3 will be using [Revolt](https://github.com/revoltphp/event-loop), I've opened a PR there that would enable the usage you were looking for here: https://github.com/revoltphp/event-loop/pull/2

PR #356, opened 2021-07-16
Add Deferred::isResolved()

trowski:


Bilge:
What's the reason for adding this now, if we've survived without it for so long?

trowski:
It's also been a pain for a long time. We added the method in v3 because we decided it was helpful, I want to back-port that to v2.

Bilge:
Do you have an example use-case? When do you need to know if it has resolved without blindly attaching a listener instead?

trowski:
A set of Deferreds for expected requests. When a request comes in the deferred is resolved with the promise for that request. If the request doesn't arrive in a particular amount of time, I want to fail the deferred with a timeout exception.

A listener on the promise doesn't quite fulfill my needs, because the request may have arrived but has not finished processing. There's not a good way to differentiate between the promise pending because the request is processing or pending because the deferred hasn't been resolved yet.

Bilge:
Couldn't such a differentiation be made by adding appropriate flags in userland code, obviating the need for this change to the library?

trowski:
Yes… and I've done that many, many times. So many times that @kelunik and I decided that it made sense to add it to the library.

kelunik (dismissed):

kelunik (approved):

PR #365, opened 2021-10-24
Implement Future::then()

trowski:
For feedback on using `then` instead of `apply`, as more people are familiar with the `then` API from JS.

### DIFF
--- a/lib/Future.php
+++ b/lib/Future.php
@@ -120,21 +120,27 @@ final class Future
      *
      * @template Tr
      *
-     * @param callable(T):Tr $onComplete
+     * @param null|callable(T):Tr $onComplete
+     * @param null|callable(\Throwable):Tr $onError
      * @return Future
      */
-    public function apply(callable $onComplete): self
+    public function then(callable $onComplete = null, callable $onError = null): self
     {
         $state = new FutureState();
 
-        $this->state->subscribe(static function (?\Throwable $error, mixed $value) use ($state, $onComplete): void {
-            if ($error) {
-                $state->error($error);
-                return;
-            }
-
+        $this->state->subscribe(static function (?\Throwable $error, mixed $value) use (
+            $state,
+            $onComplete,
+            $onError
+        ): void {
             try {
-                $state->complete($onComplete($value));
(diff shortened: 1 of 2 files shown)


xpader:
Why hasn't this been merged into V3 yet? Any concerns?

kelunik:
Because we're unsure whether we want to. `then()` brings specific expectations that we might not want to support, because it's simply not needed with fibers, e.g. returning another `Future` from inside the `then()` callback.

xpader:
I think we just need a simple `then()` at this time, not full Promise/A+ implement, like just `apply()` rename to `then()`, because name `apply` is ugly..

kelunik:
Do you have any name suggestions apart from then()? If it's not Promises/A+, I'd like to avoid that name.

xpader:
May be `done`? Still think `then` is the best name

kelunik:
Another name I had in mind was `map` like Java's optional.

xpader:
Future is a thing that can set a success and fail callback to, I think the method name must be related to success-failure, or resolved-rejected state. Name `apply` and `map` looks like have no help for understanding success or failure state.

kelunik:
`then` doesn't tell anything about success or failure either.

xpader:
`then` does not tell about sucess or failure because `then` has two parameters that include success and failure, and `then` is Promise/A+ 😓..

xpader:
It seems that I cannot convince you to change the name of apply().
I hope that the V3 API can stabilize as soon as possible so that I can develop the next version of my framework (It's alreay adapter the old V3).

kelunik:
Yeah, we've decided to avoid `then()` for now, as there's no good reason to have it other than JS having it.

xpader:
Cry.

trowski:
map is used for similar functions on placeholders in other languages. Don't get hung up on the name "then" just because JS uses it. Most guides recommend against using the second parameter (e.g., https://exploringjs.com/es6/ch_promises.html#sec_common-promise-chaining-mistakes) and use another catch call instead, so why repeat the same API mistakes here?

Issue #364, opened 2021-10-21
No callback in v3 with revolt Future?


xpader:
In some cases, we need use callback like `$promise->onResolve()`, but in v3 with revolt, no Promise now, only Future, and can not found any way to do callback?

And there is no async, asyncCallback now?

use `fn() => coroutine($callback);` instead asyncCallback?

How do we do this like before, the FutureState is internal:
```
$defer = new Defered();

SomeOtherAsyncLib::afterJobIsDone(function() use ($defer) {
    //some code before
    $defer->resolve($result);
    //some code after
});

return $defer->promise();
```

xpader:
What is the status of `then()`? Can it be merged into V3?

kelunik:
We'll use `map` as method name and keep `then()` available for potential future use.

PR #368, opened 2021-10-26
Future and Defered's complete() should have default value to null

xpader:
Sometimes we just need resolve state, don't need any return value.

https://github.com/amphp/amp/issues/364
https://github.com/amphp/amp/pull/367

### DIFF
--- a/lib/Deferred.php
+++ b/lib/Deferred.php
@@ -24,7 +24,7 @@ final class Deferred
      *
      * @param T $result Result of the operation.
      */
-    public function complete(mixed $result=null): void
+    public function complete(mixed $result = null): void
     {
         $this->state->complete($result);
     }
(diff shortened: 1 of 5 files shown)


lib/Deferred.php
```diff
@@ -24,7 +24,7 @@ public function __construct()
      *
      * @param T $result Result of the operation.
      */
-    public function complete(mixed $result): void
+    public function complete(mixed $result=null): void
```
kelunik, line 27:
```suggestion
    public function complete(mixed $result = null): void
```

lib/Future.php
```diff
@@ -65,7 +65,7 @@ public static function iterate(iterable $futures, ?CancellationToken $token = nu
      *
      * @return Future<Tv>
      */
-    public static function complete(mixed $result): self
+    public static function complete(mixed $result=null): self
```
kelunik, line 68:
```suggestion
    public static function complete(mixed $result = null): self
```

kelunik (approved):

kelunik:
Thanks!

PR #371, opened 2021-11-26
Add psalm and GitHub Actions configs to .gitattributes

sidz:
This PR:
- [x] Adds psalm config files to the .gitattributes section
- [x] Adds GitHub Actions config files to the .gitattributes section

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,10 @@
-/.* export-ignore # exclude all dotted files as mostly they are configs for other tools
-docs export-ignore
-examples export-ignore
-test export-ignore
-travis export-ignore
-phpunit.xml.dist export-ignore
-psalm.examples.xml export-ignore
-psalm.xml export-ignore
-README.md export-ignore
-CONTRIBUTING.md export-ignore
-CHANGELOG.md export-ignore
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -2,13 +2,10 @@ docs export-ignore
 examples export-ignore
 test export-ignore
 travis export-ignore
-.gitattributes export-ignore
-.gitignore export-ignore
-.travis.yml export-ignore
+/.* export-ignore
```
kelunik, line 5: It's probably a good idea to always ignore dotfiles, yeah, could you move that to the top with a comment?

sidz: done

kelunik (dismissed):

kelunik (approved):

sidz:
looks like previous variant of comment in gitattributes is not correct as github highlight it.
docs: https://git-scm.com/docs/gitignore#_pattern_format

kelunik (approved):

kelunik:
Good point, needs a fix in `byte-stream` as well then.

kelunik:
Thanks!

PR #379, opened 2021-12-04
Add variadic args to async

trowski:
We had the ability to provide function args to `Amp\async()` in the past, but removed that functionality. With first-class callables I think this would be useful again, as well as mirroring the API of `Revolt\EventLoop::queue()`.

```php
$future = Amp\async($object->method(...), $param1, $param2);
// vs.
$future = Amp\async(static fn () => $object->method($param1, $param2));
```

Similar overall, though an another benefit of the former is that it does not require the additional function call of the latter.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -13,16 +13,17 @@ use Revolt\EventLoop\UnsupportedFeatureException;
  * @template T
  *
  * @param \Closure():T $closure
+ * @param mixed ...$args Arguments forwarded to the closure when starting the fiber.
  *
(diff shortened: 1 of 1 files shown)


kelunik (approved):

xpader:
Why first argument is Closure, not callable?

kelunik:
Because anything using first-class callable syntax in PHP 8.1 will be a closure anyway. `callable` is a problematic type, because it is scope dependent, and thus e.g. not supported as property type.

Issue #381, opened 2022-01-03
Thoughts on v3


withinboredom:
Hello, I'm not sure the best place to put this... but I've been building a RAFT implementation using this library and the HTTP client/server, just for fun. I originally started with a pure fiber implementation to get to know fibers and then discovered the in-progress v3 branches when I realized just how much I'd have to implement from scratch. There's a few bits of feedback I think I'll leave here:

## concurrent awaits

A pattern I've run into is where you want to send a number of HTTP requests in parallel to multiple servers. The code ends up looking something like this:

```php
$requests = [];
foreach($servers as $server) $requests[] = async(doRequest(...), $arg1, $arg2);
foreach(Future::iterate($requests) as $next) $next->await();
```

I may have missed how to do this better, but it'd be great if this pattern could be simplified, maybe something like:

```php
// asyncParallel makes the array's value the first parameter to the callback, illustrated below:
$results = asyncParallel($servers, fn($server, $arg1, $arg2) => doRequest($server, $arg1, $arg2), $arg1, $arg2)
```

## delay/defer/repeat

There's already a `delay()` that is quite handy. `async()` appears to run before the current iteration of the event loop ends. It'd be nice to also have a `defer()` which starts on the next iteration. For example, when you know there is no point in running the callback during the same iteration, or you intentionally want to wait for a loop to let other things process. ie, semantics of `EventLoop::defer()`

Additionally, the only way to do something on a timer appears to be something like:

```php
async(function() {
  while(true) {
    delay($time);
    doWork();
  }
});
```

or use the raw `EventLoop`. It'd be great to have a `repeat()` or `schedule()` function to take advantage of the `EventLoop`'s implementation.

## Cancellations

I love this concept you've implemented here and I've created a couple of classes. I'm happy to open a PR to add them to the library:

1. ManualCancellation: allows canceling manually instead of via timeout.
2. TrappedCancellation: traps an OS signal and only cancels on that signal.
3. ScopedCancellation: a bit of a hack, but cancels once the method/function that creates the cancellation returns.

## Anything else?

I'd just like to say that I love the work y'all have done so far with fibers, it really changes the feeling of the code and makes it so much more readable. I really don't want to create any scope creep because the entire project is so huge (there's so much to do)! I just wanted to leave some feedback after trying out what I think amounts to a somewhat experimental branch. I really appreciate the work and time you've put into this library!

kelunik:
[...]

PR #382, opened 2022-01-10
Cancel DeferredCancellation when destroyed

trowski:
This will automatically cancel the associated `Cancellation` when the `DeferredCancellation` object is destroyed.

An exception object is only created when needed to avoid creating a backtrace on the normal case when the object is destroyed after all listeners have unsubscribed because the operation completed.

Related to issue #381.

### DIFF
--- a/src/DeferredCancellation.php
+++ b/src/DeferredCancellation.php
@@ -46,9 +46,19 @@ final class DeferredCancellation
         $this->cancellation = new Internal\WrappedCancellation($this->source);
     }
 
+    public function __destruct()
+    {
+        $this->source->cancel();
+    }
+
     public function getCancellation(): Cancellation
     {
         return $this->cancellation;
+    }
+
+    public function isCancelled(): bool
+    {
+        return $this->source->isRequested();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


bwoebi:
The drawback of this obviously is that the exception stacktrace is not correct when using throwIfRequested or subscribing after the cancellation.

I wonder whether we can weakreference the cancellation instead? (i.e. create in getCancellation if not existing)
Then, if the weakreference does not exist, i.e. the cancellation object is referenced nowhere else, there's no exception created.

I don't know how important the correct stacktrace is, but if it's possible without much drawback, then I'd prefer having it.
Obviously, in case the getCancellation() itself only happens after cancelling, there's not much left we can do...

kelunik:
How much does the stack trace matter if it only leads to the `__destruct` anyway?

bwoebi:
If this happens in a gc destruct, yes, then it's pretty worthless :-) Hence I'm asking how important it is.

kelunik:
I don't see how your weak reference idea should work. getCancellation should always be called on a DeferredCancellation object.

bwoebi:
That is right … but if the Cancellation object is out of scope (and destroyed before the deferred cancellation in cycle collector because newer) - that can be checked for and the exception object then not created.

kelunik (approved):

PR #383, opened 2022-01-14
Rename combinators, introduce CompositeLengthException

kelunik:


src/Future/functions.php
```diff
@@ -29,39 +29,82 @@ function race(iterable $futures, ?Cancellation $cancellation = null): mixed
...
  * @param iterable<Tk, Future<Tv>> $futures
- * @param Cancellation|null        $cancellation Optional cancellation.
+ * @param Cancellation|null $cancellation Optional cancellation.
  *
  * @return non-empty-array<Tk, Tv>
  *
  * @throws CompositeException If all futures errored.
  */
-function some(iterable $futures, int $count, ?Cancellation $cancellation = null): array
+function awaitN(iterable $futures, int $count, ?Cancellation $cancellation = null): array
```
kelunik, line 107: Should we switch the parameter order here, so `$count` will be the first param?

src/Future/functions.php
```diff
@@ -81,7 +124,7 @@ function some(iterable $futures, int $count, ?Cancellation $cancellation = null)
         }
     }
 
-    if (empty($errors)) {
+    if (\count($values) + \count($errors) < $count) {
```
kelunik, line 127: I'm not sure whether this should really be an `\Error`, IMO this should probably be an exception, as the consumer might not know how many elements are in the iterator.

bwoebi: I agree - an Exception is better.

src/Future/functions.php
trowski: Agreed.

kelunik: Should this be named `awaitAnyN`?

bwoebi: possibly there should be awaitN (= after N futures completed, we're done) and awaitAnyN (= after any N futures succeeded, we're done)?

kelunik: Yeah, discussed this with @trowski privately as well, we could introduce `awaitFirstN` later if needed.

src/Future/functions.php
kelunik: Which exception should we throw here? `CompositeException` currently requires a non-empty list of exceptions, but I'm not sure whether that's a good idea, as we could use it here without having any errors if there aren't any futures in the iterable.

trowski: Seems fine to allow an empty list.

trowski (approved):

PR #385, opened 2022-03-25
Add ClosureQueue

trowski:
Adds a simple object implementing a list of callbacks to be invoked when an event occurs.

See https://github.com/amphp/byte-stream/pull/97

### DIFF
--- a/src/ClosureQueue.php
+++ b/src/ClosureQueue.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Amp;
+
+use Revolt\EventLoop;
+
+final class ClosureQueue
+{
+    /** @var list<\Closure():void>|null */
+    private ?array $closures = [];
+
+    public function __destruct()
+    {
+        if ($this->closures !== null) {
+            $this->call();
+        }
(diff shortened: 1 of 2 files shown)


bwoebi:
What exactly is this solving which a simple Future does not solve? Create it … then subscribe on it … and you'll get notified when it finishes.

trowski:
Reusing Future for the same purpose can certainly work. Futures would be creating a number of extra objects, so there is a performance penalty. However the initial use-case for this was close notifications for streams, which isn't a particularly hot code path.

It's not uncommon for us to create lists of callbacks, though for this to be more reusable it should probably allow adding and removing callbacks and drop the destructor. Then it could be used within Future and Cancellable.

trowski:
`Future::finally()` fits the bill since it doesn't pass the resolution value to the callback.

Closing this for now. On a second look the places where a list of callbacks is used have unique parameters passed to the callback, and storing these in another object will create circular references. Making a general purpose object is maybe then more trouble than it's worth for such a simple underlying structure.

PR #390, opened 2022-05-26
Add SignalCancellation

trowski:
Used something like this a couple times, so I thought it might be worth adding.

### DIFF
--- a/src/SignalCancellation.php
+++ b/src/SignalCancellation.php
@@ -5,39 +5,49 @@ namespace Amp;
 use Revolt\EventLoop;
 
 /**
- * A TimeoutCancellation automatically requests cancellation after the timeout has elapsed.
+ * A SignalCancellation automatically requests cancellation when a given signal is received.
  */
-final class TimeoutCancellation implements Cancellation
+final class SignalCancellation implements Cancellation
 {
     use ForbidCloning;
     use ForbidSerialization;
 
-    private readonly string $watcher;
+    /** @var list<string> */
+    private readonly array $watchers;
 
     private readonly Cancellation $cancellation;
 
     /**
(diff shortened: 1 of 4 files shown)


src/SignalCancellation.php
```diff
@@ -0,0 +1,82 @@
...
+        \assert((bool) ($trace = \debug_backtrace(0)));
+
+        $callback = static function () use ($source, $message, $trace): void {
+            if ($trace) {
+                $message .= \sprintf("\r\n%s was created here: %s", self::class, Internal\formatStacktrace($trace));
+            } else {
+                $message .= \sprintf(" (Enable assertions for a backtrace of the %s creation)", self::class);
+            }
+
+            $source->cancel(new SignalException($message));
```
kelunik, line 42: What should happen if multiple signals are received?

trowski: The cancellation is triggered only on the first signal - subsequent signals are ignored. Updated to explicitly cancel all the watchers on the first signal.

src/SignalException.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Amp;
+
+/**
+ * Used as the previous exception to {@see CancelledException} when a {@see SignalCancellation} is triggered.
+ *
+ * @see SignalCancellation
+ */
+class SignalException extends \Exception
```
kelunik, line 10: Should we have another exception for that?

trowski: It can be nice if used with CombinedCancellation to differentiate which cancellation was used. It's not strictly necessary, though mirrors TimeoutException with TimeoutCancellation.

kelunik (approved):

PR #393, opened 2022-06-24
Fix `weakClosure()` not preserving original scope

Nevay:
Changes `weakClosure()` to preserve original closure scope by using `::bindTo()` instead of `::call()` if scope differs from `$this` class.


trowski:
Thanks for finding and fixing this!

PR #395, opened 2022-07-31
Allow GC of arguments during `async()` closure execution

Nevay:
Allows users to free memory if an argument is no longer required.
Requires https://github.com/revoltphp/event-loop/pull/60.


trowski:
@Nevay Thanks for the fix here as well!

PR #401, opened 2022-10-05
Declare incompatibility with revolt 0.1.x.

Bilge:
v3 is no longer compatible with `^0.1` because we rely on calls to `Revolt\EventLoop::getSuspension()` that does not exist in that version.


kelunik:
Thanks!

PR #411, opened 2023-01-08
URI example fix

Jampire:
Hi
Small fix in documentation. I'm getting following error w/o this fix: `PHP Fatal error:  Uncaught TypeError: Array must only contain instances of Amp\Future in vendor/amphp/amp/src/Future.php:36`.


kelunik (approved):

kelunik:
Thanks!

PR #420, opened 2023-07-24
Add spawn function

danog:


kelunik:
As previously discussed, I don't see any benefit over using async().

danog:
One very clear benefit is to avoid the overhead of creating a future, without having to use `EventLoop::queue`.

trowski:
The main purpose I see to this is providing a tool to create a fiber without the future within the Amp eco-system. Otherwise obviously it is exactly the same as using `EventLoop::queue()`. I can see benefits to this, as it might be easier for newcomers to discover `Amp\spawn()` over needing to read the event loop docs.

For a full compliment of `Amp` namespaced functions, we'd probably also need `Amp\setEventLoopErrorHandler()` and `Amp\runEventLoop()` (or similar). Maybe I've forgotten something? I think without these as well, there's not much benefit because users will still need to learn about the event loop API.

PR #421, opened 2023-07-25
Allow referencing a TimeoutCancellation

danog:


src/TimeoutCancellation.php
```diff
@@ -17,10 +17,11 @@ final class TimeoutCancellation implements Cancellation
...
 
     /**
-     * @param float  $timeout Seconds until cancellation is requested.
-     * @param string $message Message for TimeoutException. Default is "Operation timed out".
+     * @param float  $timeout     Seconds until cancellation is requested.
+     * @param string $message     Message for TimeoutException. Default is "Operation timed out".
+     * @param bool   $unreference Whether to unreference the timer.
      */
-    public function __construct(float $timeout, string $message = "Operation timed out")
+    public function __construct(float $timeout, string $message = "Operation timed out", bool $unreference = true)
```
kelunik: I really dislike any boolean parameter that's not purely a setter. Should probably be a separate method?

kelunik:
What's the use case?

src/TimeoutCancellation.php
trowski: We did use a similar parameter in `delay`. IMO, if we have a boolean parameter here it should match `delay`, being called `$reference` and default to false.

trowski:
I think reference and unreference methods similar to Socket would make sense on `TimeoutCancellation` and `SignalCancellation`. Perhaps no constructor parameter at all then?

trowski:
For anyone landing here wondering why the commit referenced above was reverted: We decided that the use-case for `TimeoutCancellation` being referenced was an abuse of the API, and more appropriate APIs such as `Amp\delay()` should have been used instead.

PR #430, opened 2024-02-20
Update functions.php

lav45:
Skip delay if timeout is equels 0.0

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -62,6 +62,9 @@ function now(): float
  */
 function delay(float $timeout, bool $reference = true, ?Cancellation $cancellation = null): void
 {
+    if ($timeout === 0.0) {
+        return;
+    }
     $suspension = EventLoop::getSuspension();
     $callbackId = EventLoop::delay($timeout, static fn () => $suspension->resume());
     $cancellationId = $cancellation?->subscribe(


bwoebi:
`\Amp\delay(0);` is an intentional feature to force the fiber to continue in the next event-loop tick.

So, no, we don't want to skip this. ... Especially not in 3.x, because it would be a breaking change :-/

Even though I agree, a separate function for "schedule in next tick" rather than \Amp\delay(0) would have been good, but too late now.

trowski:
@bwoebi  FYI, `Amp\async(fn () => doSomething())` effectively will schedule something in the next tick. `Amp\delay(0)` is a nice way of pausing the current fiber for a tick, something which I generally only need to do in tests anyway.

bwoebi:
@trowski async runs after all current outstanding microtasks, not next event-loop tick. `awaitOutstandingMicrotasks()` and `awaitEventloopTick()` would be sort of nice to have sometimes, explicitly.

trowski:
Yes, though I find it rare to need to make such a distinction. I suppose if you're dealing with event-loop callback timing in a test that timing is going to matter. In that case, there's also `EventLoop::defer()`, which I guess is actually what I should have suggested over `Amp\async()`.

xpader:
I think there is no reason to change this.

PR #441, opened 2024-07-24
Add Interval

trowski:
This adds a utility class `Interval` which controls a repeating timer, automatically cancelling the timer when the object is destroyed.

The main use-case of this object is to store an instance of `Interval` as an object property. When the holding object is destroyed, the repeating timer will be cancelled. This avoids a bit of boilerplate in the class, such as needing to declare a destructor to manually cancel the event loop callback.

The value here isn't necessarily huge, but this is a pattern we frequently use for classes with self-cleanup mechanisms such as [`LocalCache`](https://github.com/amphp/cache/blob/46912e387e6aa94933b61ea1ead9cf7540b7797c/src/LocalCache.php).

### DIFF
--- a/src/Interval.php
+++ b/src/Interval.php
@@ -10,7 +10,7 @@ use Revolt\EventLoop;
  */
 final class Interval
 {
-    private readonly string $watcher;
+    private readonly string $callbackId;
 
     /**
      * @param float $interval Invoke the function every $interval seconds.
@@ -20,16 +20,16 @@ final class Interval
      */
     public function __construct(float $interval, \Closure $closure, bool $reference = true)
     {
-        $this->watcher = EventLoop::repeat($interval, $closure);
+        $this->callbackId = EventLoop::repeat($interval, $closure);
 
         if (!$reference) {
-            EventLoop::unreference($this->watcher);
+            EventLoop::unreference($this->callbackId);
         }
     }
 
     public function __destruct()
     {
-        EventLoop::cancel($this->watcher);
+        EventLoop::cancel($this->callbackId);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Interval.php
```diff
@@ -0,0 +1,98 @@
...
+
+use Revolt\EventLoop;
+
+/**
+ * This object invokes the given callback within a new coroutine every $interval seconds until either the
+ * {@see self::disable()} method is called or the object is destroyed.
+ */
+final class Interval
+{
+    private readonly string $watcher;
```
kelunik:
```suggestion
    private readonly string $callbackId;
```

src/Interval.php
```diff
@@ -0,0 +1,98 @@
...
+
+        if (!$reference) {
+            EventLoop::unreference($this->watcher);
+        }
+    }
+
+    public function __destruct()
+    {
+        EventLoop::cancel($this->watcher);
+    }
```
kelunik: Shouldn't there also be a manual `cancel` method?

kelunik (changes requested):

src/Interval.php
trowski: I think it makes more sense for the object to represent a valid repeating timer. If we allow external cancellation, then the object needs to expose an `isCancelled()` method so the state isn't hidden.

xpader:
It's just a `auto destruct` for EventLoop::repeat()??
I think it's not very necessary for a base library.

trowski:
How about moving this to `amphp/sync`? That library has sort of become a utility library.

kelunik:
I'd rather keep this here, as it's very basic, no?

trowski:
@kelunik Yeah, I can't see this needing changing, so seems basic enough to go here. Should we add this then?

PR #450, opened 2025-01-10
Supplement the missing parts of runnable code

lwlwilliam:
I found that there are not many sample codes for amphp and other library documents under the organization, and there are various problems with the sample codes, some without `<?php` tags, some without require autoload.chp, and some without use statements. I'm not sure if I need help refining these examples so that they can run directly.


trowski:
Thanks! Our code snippets could use some love, and this is a good first step.

PR #451, opened 2025-05-01
Replace \Closure(...) with \Closure(mixed...)

vudaltsov:
See https://github.com/phpstan/phpstan/issues/12954

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,8 +447,8 @@ time differences independent of wall-time.
 ```php
 /**
  * @template TReturn
- * @param Closure(mixed...):TReturn $closure
- * @return Closure(mixed...):TReturn
+ * @param Closure(...):TReturn $closure
+ * @return Closure(...):TReturn
  */
 function weakClosure(Closure $closure): Closure
 ```
--- a/README.md
+++ b/README.md
@@ -447,8 +447,8 @@ time differences independent of wall-time.
 ```php
 /**
  * @template TReturn
- * @param Closure(...):TReturn $closure
- * @return Closure(...):TReturn
+ * @param Closure(mixed...):TReturn $closure
+ * @return Closure(mixed...):TReturn
  */
 function weakClosure(Closure $closure): Closure
 ```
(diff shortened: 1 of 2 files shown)


vudaltsov:
The `\Closure(...)` is not documented, by the way. This [page](https://psalm.dev/docs/annotating_code/type_syntax/callable_types/) only says that a type can be **suffixed** with `...`.

trowski:
I originally removed this syntax to support PHPStan: https://github.com/amphp/amp/commit/138801fb68cfc9c329da8a7b39d01ce7291ee4b0

So this has now changed?

vudaltsov:
🤦🤦🤦 

Well, the only thing I can tell is that with `mixed...` it now works in both analyzers:
https://phpstan.org/r/681abd9e-ec0f-4edb-ab82-134e3be7088e
https://psalm.dev/r/2dd7ca2ae0

vudaltsov:
In fact my solution is wrong: https://phpstan.org/r/9c17bb5f-4087-4201-b684-72a28b0e069c . 

PHPStan is correct: you cannot pass a `Closure` with narrower parameters to a `Closure` with 
wider parameters, since parameters are contravariant.

This is how `async` function would [look like in TypeScript](https://www.typescriptlang.org/play/?target=99#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXygFsAHAQQGcBPdAHlJgHNz4QAPDEVYZqVSgbQC6AGngAVAHwAKAFDx58SDnLI4ALnhSAdDthMN9JgEp4AXgnjhchTq17yBxuRlGNYgNwyZRMlXRSpVg1UZEIAIxAYV3hyDBgsVAYzCwByWPjElNEARiNvEgpqMACgmLiEhmj0iuT4NPLMnKMgA) (thanks to @klimick for helping with the snippet). @ondrejmirtes is already discussing smth similar: https://github.com/phpstan/phpstan/issues/8214.

Here's the possible solution:
- for PHPStan use a `Closure` with a bunch of `never` parameters for  to indicate that `Closure` might have parameters with any wider types and suppress the place where it is invoked,
- for Psalm keep the current `Closure(...)` syntax, because Psalm does not allow to have `never` parameters.

https://phpstan.org/r/4e98ea03-a863-4b39-8fcf-07cfdc2e8ea4
https://psalm.dev/r/cda9b2565b

PR #456, opened 2025-08-27
Use str_increment() to fix PHP 8.5 deprecation

nicolas-grekas:


kelunik:
@nicolas-grekas Can you please have a look at the build failure? I've already pushed a change that changes the action versions.

src/CompositeCancellation.php
```diff
@@ -69,7 +69,8 @@ public function __destruct()
 
     public function subscribe(\Closure $callback): string
     {
-        $id = $this->nextId++;
+        $id = $this->nextId;
+        \PHP_VERSION_ID >= 80300 ? $this->nextId = \str_increment($this->nextId) : ++$this->nextId;
```
xabbuh, line 73:
```suggestion
        /** @psalm-suppress UndefinedFunction */
        \PHP_VERSION_ID >= 80300 ? $this->nextId = \str_increment($this->nextId) : ++$this->nextId;
```

Probably like this (as in revoltphp/event-loop#110)?

kelunik: I disabled it for that function globally instead.

PR #457, opened 2025-09-03
Use str_increment() to fix PHP 8.5 deprecation

nicolas-grekas:
Same as #456 for branch 2.x, which is still in use.

Also updates the GitHub action versions to fix the build.

### DIFF
--- a/test/InvalidYieldErrorTest.php
+++ b/test/InvalidYieldErrorTest.php
@@ -33,6 +33,6 @@ class InvalidYieldErrorTest extends BaseTest
         })();
 
         $error = new InvalidYieldError($gen, "prefix");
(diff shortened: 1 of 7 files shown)


test/InvalidYieldErrorTest.php
```diff
@@ -33,6 +33,6 @@ public function testSubgenerator()
         })();
 
         $error = new InvalidYieldError($gen, "prefix");
-        $this->assertSame("prefix; integer yielded at key 'foo' on line " . (__LINE__ - 8) . " in " . __FILE__, $error->getMessage());
+        $this->assertSame("prefix; integer yielded at key 'foo' on line " . (__LINE__ - 8 - (int) (\PHP_VERSION_ID >= 80400)) . " in " . __FILE__, $error->getMessage());
```
kelunik, line 36: Interesting, that the line was always wrong.

kelunik:
Thanks!

PR #460, opened 2026-03-26
Amp\disperse() function

rela589n:
In this pull request, I want to introduce a `disperse($closures)` function.

It is created to asynchronously execute an array of closures when the client needs all results.

```php
$results = \Amp\disperse([
    fn () => $httpClient->request(new Request('https://www.google.com', 'HEAD')),
    fn () => $httpClient->request(new Request('https://www.bing.com', 'HEAD')),
]);
```

Initially,  I considered placing this function in `Amp\Future` namespace, but later I opted in favor of `Amp`, since this function doesn't have Futures in the signature.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -5,6 +5,8 @@ namespace Amp;
 use Revolt\EventLoop;
 use Revolt\EventLoop\UnsupportedFeatureException;
 
+use function Amp\Future\awaitAll;
+
 /**
  * Creates a new fiber to execute the given closure asynchronously. A Future is returned which is completed with the
  * return value of the passed closure or will fail if the closure throws an exception.
@@ -41,6 +43,36 @@ function async(\Closure $closure, mixed ...$args): Future
     EventLoop::queue($run, $state, $closure, $args);
 
     return new Future($state);
+}
+
+/**
+ * Executes the given closures concurrently and returns their results as soon as all closures complete successfully.
+ *
+ * @template Tk of array-key
+ * @template Tv
+ *
+ * @param array<Tk, \Closure():Tv> $closures
+ * @param Cancellation|null $cancellation Optional cancellation.
+ *
+ * @return array<Tk, Tv> Results in input order of Closures.
+ */
+function disperse(array $closures, ?Cancellation $cancellation = null): array
+{
+    if ([] === $closures) {
+        return [];
+    }
+
+    $futures = \array_map(static fn (\Closure $closure): Future => async($closure), $closures);
+
+    [$errors, $values] = awaitAll($futures, $cancellation);
+
+    if ($errors) {
+        /** @var non-empty-array<array-key, \Throwable> $errors */
+        throw new CompositeException($errors);
+    }
+
+    /** @var non-empty-array<Tk, Tv> */
+    return \array_replace($closures, $values);
 }
 
 /**
--- a/README.md
+++ b/README.md
@@ -278,6 +278,17 @@ individual keys in the component array are preserved from the `iterable` passed 
 
 `Amp\Future\awaitAll($iterable, $cancellation)` awaits all futures and returns their results as `[$errors, $values]` array.
 
+##### disperse
+
+`Amp\disperse($closures, $cancellation)` executes all given closures asynchronously and returns their results on completion.
+
+```php
+$results = \Amp\disperse([
+    fn () => $httpClient->request(new Request('https://www.google.com', 'HEAD')),
+    fn () => $httpClient->request(new Request('https://www.bing.com', 'HEAD')),
+]);
+```
+
 ##### awaitFirst
 
 `Amp\Future\awaitFirst($iterable, $cancellation)` unwraps the first completed `Future`, whether successfully completed or errored.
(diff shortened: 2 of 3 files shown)


src/functions.php
```diff
@@ -43,6 +45,37 @@ function async(\Closure $closure, mixed ...$args): Future
     return new Future($state);
 }
 
+/**
+ * Executes the given closures concurrently and returns their results as soon as all closures complete successfully.
+ *
+ * The returned array preserves the completion order of the closures, not the input order.
```
kelunik: Why did you choose completion order rather than input order?

xpader:
What is that??? `await(iterable $futures)` race order version ???
I don't think we need add this to amp library.

src/functions.php
xpader: Not a good idea.

rela589n: @kelunik , you are right to question that. I primarily thought about array de-structuring (which works well by the preserved indices), but really, call_user_func will confuse the order.

rela589n:
@xpader , if you don't find this useful at all, you are the contributor, and I'm a layman - it's in your authority to close.

rela589n:
Maybe this's not the best name for such a function.
Maybe naming it some other way would bring clarity?
What's on your mind?

rela589n:
Hi, @webpatser, do you think it's a good thing to include?

rela589n:
Hi, guys
Can we move toward some logical conclusion about this?

rela589n:
Hi, @xpader !
How is it going?

xpader:
@rela589n I don't thing we shoud add this to amp library, you can use it yourself.

rela589n:
I hope you're doing well.

Of course, I could have added this to my repo; I would not have opened the PR then.

Yet, I think you wouldn't want to bother yourself with `async()` and `await()` functions when you just want to run something "in parallel".

For example, consider a cache warmup scenario with `disperse`:

```php
disperse($cacheWarmupFunctions);
```

Without `disperse` we'd have to wrap those into async/await:

```php
$futures = array_map(async(...), $cacheWarmupFunctions);
awaitAll($futures);
```

Which obscures the high picture.

I don't think you'd want to write this every time if you can avoid writing this all the time.

webpatser:
Hey @rela589n, sorry for the slow turnaround, your ping slipped past me. Cheers @xpader for chiming in.

My honest take: I don't think `disperse()` earns a spot in core. The wrapper is two lines today:

```php
Future\await(array_map(async(...), $closures));
```

That already reads cleanly once `async()`/`await()` are familiar, and a third name for the same thing grows the API surface without unlocking new behavior. The naming is also tricky; "disperse" didn't immediately read as "run all in parallel and collect results" to me. `awaitAll` or `parallel` are closer, but both come with their own history.

The cache-warmup example is fair. That's a great fit for a small helper in app code or a userland package, where you can name it whatever fits your team.

Not a hard no. If the maintainers want it in, I won't push back. But my vote leans toward closing and keeping the helper in user land.

kelunik:
I think there's value in having such a function, you don't want to write that array_map repeatedly. I'm however not sure on the name:

> Maybe this's not the best name for such a function.
> Maybe naming it some other way would bring clarity?

There likely is a better name. And we should be really sure it's a good name if we add it to the core library, because we can't just release another major version of it if we're wrong.

I'm totally happy to have this sit around for a while, if someone comes good with naming suggestions and it's still good after a while, we can proceed and merge this.

rela589n:
What do you think about "scatter"? 

It means: (of a group of people or animals) separate and move off quickly in different directions.

kelunik:
I think it needs to somehow cover the dispersion and then collecting the results again, thought about `forkJoin` (but don't like it because fork usually refers to processes) and `runConcurrently` so far.

rela589n:
If we consider forkJoin, at first, I didn't really think of "fork" as of a verb, but of a noun.

> cover the dispersion and then collecting the results again

TBH, it reminds me of [Scatter-Gather](https://www.enterpriseintegrationpatterns.com/patterns/messaging/BroadcastAggregate.html) (Not that I ever used it, but it clicked in me) enterprise pattern: 

> broadcasts a message to multiple recipients and re-aggregates the responses back into a single message.

In our case, it is:

```php
$results = \Amp\scatterGather([
    fn () => $httpClient->request(new Request('https://www.google.com', 'HEAD')),
    fn () => $httpClient->request(new Request('https://www.bing.com', 'HEAD')),
]);
```

webpatser:
Hey @rela589n, @kelunik, on naming.

`scatterGather` reads as enterprise service bus to me. Accurate to the pattern, but it sets the wrong expectation for a small concurrent-execute helper.

My vote is `runConcurrently`. Says what it does, no metaphor. `Amp\runConcurrently([...])` reads fine at the call site. If the length grates, `concurrent` as a shorter alternative.

rela589n:
IMO `concurrent` is better compared to `runConcurrently`

```php
$results = \Amp\concurrent([
    fn () => $httpClient->request(new Request('https://www.google.com', 'HEAD')),
    fn () => $httpClient->request(new Request('https://www.bing.com', 'HEAD')),
]);
```

PR #468, opened 2026-06-29
Add `concurrent()` function

rela589n:
This PR adds a `concurrent()` function for creating a fixed set of concurrent operations:

```php
function concurrent(iterable<\Closure> $closures): iterable<Future>
```

When combined with other functions through the pipe operator, it produces self-explanatory concurrent code:

```php
$results = [
    fn () => $httpClient->request(new Request('https://google.com', 'HEAD')),
    fn () => $httpClient->request(new Request('https://bing.com', 'HEAD')),
] |> concurrent(...) |> await(...);
```

```php
$firstReachableApi = [
    fn () => $httpClient->request(new Request('https://a.api.com', 'HEAD')),
    fn () => $httpClient->request(new Request('https://b.api.com', 'HEAD')),
] |> concurrent(...) |> awaitAny(...);
```

The ability to pass `iterable` into `concurrent()` suits the cases in which the exact number of processing closures is unknown beforehand. Thus, one can even await an unbounded producer:

```php
\Amp\Future\awaitAll(\Amp\concurrent(server()));

/** @return iterable<\Closure> */
function server(): iterable
{
    foreach (requests() as $request) {
        yield $request->handle(...);
    }
}

/** @return iterable<Request> */
function requests(): iterable
{
    while (null !== ($request = waitForRequest())) {
        yield $request;
    }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,7 +219,22 @@ throws an exception.
 
 #### Combinators
 
-In concurrent applications, there will be multiple futures, where you might want to await them all or just the first one.
+In concurrent applications, there will be multiple futures, where you might want to await them all or just the first
+one.
+
+You can create a bunch of futures by applying `Amp\concurrent()` to an iterable of closures:
+it returns a `Future` for each, preserving the keys.
+
+```php
+$firstReachableApiResponse = \Amp\Future\awaitAny([
+    fn () => $httpClient->request(new Request('https://a.api.com', 'HEAD')),
+    fn () => $httpClient->request(new Request('https://b.api.com', 'HEAD')),
+] |> \Amp\concurrent(...));
+```
+
+`Amp\concurrent()` returns `\Generator`, so producers of unbounded length are supported.
+
+The combinators below await such futures in different ways.
 
 ##### await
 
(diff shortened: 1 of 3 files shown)


src/functions.php
```diff
@@ -43,6 +43,23 @@ function async(\Closure $closure, mixed ...$args): Future
...
+ * Pass or pipe the result into a combinator such as {@see Future\await()} to await the values.
+ *
+ * @template Tk of array-key
+ * @template Tv
+ *
+ * @param array<Tk, \Closure():Tv> $closures
+ *
+ * @return array<Tk, Future<Tv>> A Future for each closure, with keys preserved.
+ */
+function concurrent(array $closures): array
```
kelunik: Do we want to support iterable here?

rela589n: It's not a problem with me to make it `iterable` / `array`,
yet the decision is yours, guys.

rela589n: Probably `iterable` would work better.

For example, a concept of a simple async server:

```php
\Amp\Future\settle(\Amp\concurrent(server()));

/** @return iterable<\Closure> */
function server(): iterable
{
    foreach (requests() as $request) {
        yield $request->handle(...);
    }
}

/** @return iterable<Request> */
function requests(): iterable
{
    while (true) {
        yield waitForRequest();
    }
}
```

I think there might be some situations in which the exact number of processing closures is not known aforehand, so accepting `iterable` then makes sense.

test/ConcurrentTest.php
```diff
@@ -0,0 +1,92 @@
...
+                delay(0.01);
+
+                yield static fn (): int => $i;
+            }
+        })();
+
+        self::assertSame([0, 1], Future\awaitAnyN(2, concurrent($closures)));
+    }
+
+    public function testClosuresAreStartedRightOffWithAwait(): void
```
kelunik: What is this test supposed to test?

README.md
```diff
@@ -219,7 +219,23 @@ throws an exception.
...
+it returns a `Future` for each, preserving the keys.
+
+```php
+$firstReachedApi = [
+    fn () => $httpClient->request(new Request('https://a.api.com', 'HEAD')),
+    fn () => $httpClient->request(new Request('https://b.api.com', 'HEAD')),
+] |> \Amp\concurrent(...) |> \Amp\Future\awaitAny(...);
+```
+
+`Amp\concurrent()` is lazy: closures are started as the combinator consumes the returned generator, so producers
```
kelunik: While it is kind of lazy, all combinators etc. will consume the entire iterator, so this might give a wrong impression.

